PR #26, opened 2015-01-07
Remove `if condition bool else bool`

calavera:
This is a really silly change, feel free to ignore it, but I just couldn't help myself :smile:


mitchellh:
LGTM, I'll let @ryanuber merge if he feels its correct still. The only logic change I can think of is `> 0` instead of `!= 0` which I think is fine, but just want to verify.

ryanuber:
Yeah seems good to me, I doubt we'll have a negative-length slice so `> 0` is probably good :). Thanks @calavera!

PR #28, opened 2015-02-23
Custom errors for merge delegate

ryanuber:
Allows `NotifyMerge()` to return an error instead of just a bool. This makes it easier to identify why a merge is canceled.


armon:
LGTM

PR #30, opened 2015-03-08
LocalNode uses assigned port rather than the Config.BindPort

dadgar:
Create a Node using its bound port rather than the port specified in the Config. There is a distinction when the Config specifies port 0, allowing Memberlist to bind to any free port. This change does not effect the port if an AdvertiseAddr/AdvertisePort is explicitly set in the Config.


armon:
LGTM

PR #32, opened 2015-03-15
Fix race condition in Leave

ErikDubbelboer:
[Leave](https://github.com/hashicorp/memberlist/blob/50c5c11867c98113b8390564b2897edeb6c856d0/memberlist.go#L448) has several race conditions.

1. It [writes to `m.leave`](https://github.com/hashicorp/memberlist/blob/50c5c11867c98113b8390564b2897edeb6c856d0/memberlist.go#L457) while only having a [lock on `m.startStopLock`](https://github.com/hashicorp/memberlist/blob/50c5c11867c98113b8390564b2897edeb6c856d0/memberlist.go#L449). At the same time `m.leave` is read in [`aliveNode`](https://github.com/hashicorp/memberlist/blob/master/state.go#L610) and [`deadNode`](https://github.com/hashicorp/memberlist/blob/master/state.go#L861) while not holding this lock.

2. It [reads from `m.nodeMap`](https://github.com/hashicorp/memberlist/blob/50c5c11867c98113b8390564b2897edeb6c856d0/memberlist.go#L459) while not holding a lock on `m.nodeLock`.

Both these issues are fixed in the pull request by only using `m.nodeLock`.

Seeing as the test cases for memberlist are full of race conditions I don't think it's useful to write a test case for this. I found this race condition while testing a different program that uses memberlist.

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -53,8 +53,6 @@ type Memberlist struct {
 
 	broadcasts *TransmitLimitedQueue
 
-	startStopLock sync.Mutex
-
 	logger *log.Logger
 }
 
@@ -228,7 +226,6 @@ func (m *Memberlist) resolveAddr(hostStr string) ([][]byte, uint16, error) {
 // as if we received an alive notification our own network channel for
 // ourself.
 func (m *Memberlist) setAlive() error {
-
 	var advertiseAddr []byte
 	var advertisePort int
 	if m.config.AdvertiseAddr != "" {
@@ -446,10 +443,12 @@ func (m *Memberlist) NumMembers() (alive int) {
 // This method is safe to call multiple times, but must not be called
 // after the cluster is already shut down.
 func (m *Memberlist) Leave(timeout time.Duration) error {
-	m.startStopLock.Lock()
-	defer m.startStopLock.Unlock()
+	m.nodeLock.Lock()
+	// We can't defer m.nodeLock.Unlock() because m.deadNode will also try to
+	// acquire a lock so we need to Unlock before that.
 
 	if m.shutdown {
+		m.nodeLock.Unlock()
 		panic("leave after shutdown")
 	}
 
@@ -457,6 +456,7 @@ func (m *Memberlist) Leave(timeout time.Duration) error {
 		m.leave = true
 
 		state, ok := m.nodeMap[m.config.Name]
+		m.nodeLock.Unlock()
 		if !ok {
 			m.logger.Printf("[WARN] memberlist: Leave but we're not in the node map.")
 			return nil
@@ -480,6 +480,8 @@ func (m *Memberlist) Leave(timeout time.Duration) error {
 				return fmt.Errorf("timeout waiting for leave broadcast")
 			}
 		}
+	} else {
+		m.nodeLock.Unlock()
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


mikepb:
Hi @ErikDubbelboer,

I'm working on an experimental implementation of the SWIM failure detector: https://github.com/mikepb/go-swim

The library is primarily meant to complement memberlist in the research arena by decomposing SWIM into interchangeable components to evaluate changes to the algorithm via simulation.

I'm using an event loop design (as in JavaScript) to simplify concurrency control. The only explicit lock used is for sending messages with piggybacked broadcast messages. The event loop implicitly uses locks backing Go channels.

I'd love to get your thoughts on this design. Would you think using an event loop design would be more long-term maintainable?

Best,
Michael

ErikDubbelboer:
Hi @mikepb 

I looked at your project quickly and it's quite clean and easy to read code :)

I don't have too much time to look at it right now. But from my first look I can at least already tell you that I think you have the exact same race condition in your code as this pull fixes in memberlist. [`Detector.started`](https://github.com/mikepb/go-swim/blob/92136f9ee2c3f2d55ab362c8fc3f25a37f9a5b50/detector.go#L32) can be written to from [`Detector.Stop()`](https://github.com/mikepb/go-swim/blob/92136f9ee2c3f2d55ab362c8fc3f25a37f9a5b50/detector.go#L171) while it's being read from [`Detector.recv()`](https://github.com/mikepb/go-swim/blob/92136f9ee2c3f2d55ab362c8fc3f25a37f9a5b50/detector.go#L458).

Another potential issue I noticed is that because your event loop doesn't use goroutines to dispatch event to the users of your library I think you should be very careful in documenting what your users can do. For example I see your [`Detector.handle()`](https://github.com/mikepb/go-swim/blob/92136f9ee2c3f2d55ab362c8fc3f25a37f9a5b50/detector.go#L521) function puts messages in the `Detector.MessageCh` channel in a blocking way. If a user of your library doesn't use a big enough buffer for this channel and/or doesn't read from this channel often enough it would block your whole receive loop from processing any messages while waiting on this. (memberlist actually has this exact same issue and has a [specific warning for this](https://github.com/hashicorp/memberlist/blob/master/net.go#L251)).

I think the event loop design that you have chosen works pretty well for this kind functionality.

I like the way you can now simulate the network and test things like failures or delays. For the project in which I'm currently using memberlist I actually wrote [UDP](https://github.com/atomx/dcounter/blob/master/testproxy/udp.go) and [TCP](https://github.com/atomx/dcounter/blob/master/testproxy/tcp.go) proxies to [simulate network failure](https://github.com/atomx/dcounter/blob/master/server/fail_test.go#L37).

mikepb:
Thanks @ErikDubbelboer !

I was also considering whether to account for that race condition. I suppose that it would be more proper to program defensively, even if the `Start()` and `Stop()` methods are not meant to be called from multiple goroutines.

The blocking side channels should definitely have documentation describing appropriate use cases. I'm just not sure what those would be at the moment.

ErikDubbelboer:
For a good reason to merge this: https://software.intel.com/en-us/blogs/2013/01/06/benign-data-races-what-could-possibly-go-wrong

ryanuber:
This makes sense to me. I'll let @armon comment in case there is any impact he sees before merging this, but thanks @ErikDubbelboer!

armon:
Thanks!

PR #33, opened 2015-03-17
Change RTTDelegate to PingDelegate

derekchiang:


ping_delegate.go
```diff
@@ -0,0 +1,13 @@
...
+import "time"
+
+// PingDelegate is used to notify an observer how long it took for a ping message to
+// complete a round trip.  It can also be used for writing arbitrary byte slices
+// into ack messages.
+type PingDelegate interface {
+	// AckPayload is invoked when an ack is being sent; the returned bytes will be appended to the ack
+	AckPayload() []byte
+	// NotifyPing is invoked when an ack for a ping is received
+	NotifyPing(other *Node, rtt time.Duration, payload []byte)
```
armon: I would change this to `NotifyPingComplete` since it's unclear if you are being notified that a ping is received or that a ping is complete currently.

state.go
```diff
@@ -215,10 +215,14 @@ func (m *Memberlist) probeNode(node *nodeState) {
 
 	// Setup an ack handler
 	ackCh := make(chan bool, m.config.IndirectChecks+1)
-	m.setAckChannel(ping.SeqNo, ackCh, m.config.ProbeInterval)
+	sent := time.Now()
+	m.setAckChannel(ping.SeqNo, ackCh, func(payload []byte) {
```
armon: This is a bit dirty and works around the indented use of an ack channel. Lets add a new type `ackResponse` that includes the `Payload` and `Complete` bool, and change the type of the channel to `ackResponse` instead of bool. This way we aren't overloading by injecting a second handler.

armon:
@derekchiang Left some minor feedback!

derekchiang:
@armon Feedback addressed.

armon:
LGTM!

PR #37, opened 2015-06-02
Adds TCP direct pings.

slackpad:
Here's an initial cut at the fix for hashicorp/consul#742. It adds a TCP-based direct ping pipelined with the normal UDP indirect pings if the direct UDP ping fails. If all the UDP-based pings fail and the TCP-based ping works, then a log message will be generated to alert the user that their networking config isn't right.

### DIFF
--- a/net.go
+++ b/net.go
@@ -196,17 +196,19 @@ func (m *Memberlist) handleConn(conn *net.TCPConn) {
 	defer conn.Close()
 	metrics.IncrCounter([]string{"memberlist", "tcp", "accept"}, 1)
 
+	conn.SetDeadline(time.Now().Add(m.config.TCPTimeout))
 	msgType, bufConn, dec, err := m.readTCP(conn)
 	if err != nil {
 		m.logger.Printf("[ERR] memberlist: failed to receive: %s", err)
 		return
 	}
 
-	if msgType == userMsg {
+	switch msgType {
+	case userMsg:
 		if err := m.readUserMsg(bufConn, dec); err != nil {
 			m.logger.Printf("[ERR] memberlist: Failed to receive user message: %s", err)
 		}
-	} else if msgType == pushPullMsg {
+	case pushPullMsg:
 		join, remoteNodes, userState, err := m.readRemoteState(bufConn, dec)
 		if err != nil {
 			m.logger.Printf("[ERR] memberlist: Failed to read remote state: %s", err)
@@ -221,7 +223,31 @@ func (m *Memberlist) handleConn(conn *net.TCPConn) {
 		if err := m.mergeRemoteState(join, remoteNodes, userState); err != nil {
 			return
 		}
-	} else {
+	case pingMsg:
+		var p ping
+		if err := dec.Decode(&p); err != nil {
+			m.logger.Printf("[ERR] memberlist: Failed to decode TCP ping: %s", err)
+			return
+		}
+
+		if p.Node != "" && p.Node != m.config.Name {
+			m.logger.Printf("[WARN] memberlist: Got ping for unexpected node %s", p.Node)
+			return
+		}
+
+		ack := ackResp{p.SeqNo, nil}
+		out, err := encode(ackRespMsg, &ack)
+		if err != nil {
+			m.logger.Printf("[ERR] memeberlist: Failed to encode TCP ack: %s", err)
+			return
+		}
+
+		err = m.rawSendMsgTCP(conn, out.Bytes())
+		if err != nil {
+			m.logger.Printf("[ERR] memberlist: Failed to send TCP ack: %s", err)
+			return
+		}
+	default:
 		m.logger.Printf("[ERR] memberlist: Received invalid msgType (%d)", msgType)
 	}
 }
@@ -654,6 +680,7 @@ func (m *Memberlist) sendAndReceiveState(addr []byte, port uint16, join bool) ([
 		return nil, nil, err
 	}
 
+	conn.SetDeadline(time.Now().Add(m.config.TCPTimeout))
 	msgType, bufConn, dec, err := m.readTCP(conn)
 	if err != nil {
 		return nil, nil, err
@@ -789,9 +816,6 @@ func (m *Memberlist) decryptRemoteState(bufConn io.Reader) ([]byte, error) {
 // readTCP is used to read the start of a TCP stream.
 // it decrypts and decompresses the stream if necessary
 func (m *Memberlist) readTCP(conn net.Conn) (messageType, io.Reader, *codec.Decoder, error) {
-	// Setup a deadline
-	conn.SetDeadline(time.Now().Add(m.config.TCPTimeout))
-
 	// Created a buffered reader
 	var bufConn io.Reader = bufio.NewReader(conn)
 
@@ -934,7 +958,7 @@ func (m *Memberlist) mergeRemoteState(join bool, remoteNodes []pushNodeState, us
 	return nil
 }
 
-// readUserMsg is used to decode a UserMsg from a TCP stream
+// readUserMsg is used to decode a userMsg from a TCP stream
 func (m *Memberlist) readUserMsg(bufConn io.Reader, dec *codec.Decoder) error {
 	// Read the user message header
 	var header userMsgHeader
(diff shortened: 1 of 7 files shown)


slackpad:
I still need to test this integrated into Consul, and I'd also like to beef up the unit tests a little but I thought I'd put it up for some feedback.

state.go
```diff
@@ -257,6 +258,37 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
 	}
 
+	// Also make an attempt to contact the node directly over TCP. This
+	// helps prevent confused clients who get isolated from UDP traffic
+	// but can still speak TCP (which also means they can possibly report
+	// misinformation to other nodes via anti-entropy), avoiding flapping in
+	// the cluster.
+	fallbackCh := make(chan bool)
+	if node.PMax >= 3 {
+		destAddr := &net.TCPAddr{IP: node.Addr, Port: int(node.Port)}
```
slackpad: It's a little dirty to have this net stuff up here, but I want to silently move on if the connect fails. If this is down inside sendPingAndWaitForAck then it complicates the return values (you need to be able to say "there were no errors but we wen't able to get a ping and an ack".

armon: I think we can move this all into `sendPingAndWaitForAck` and change the return to a `(bool, error)`. It should be enough to have the bool indicated `didContact`

armon:
@slackpad This looks SUPER solid.  My major feedback is I think we can simplify the state change and push more down into `sendPingAndWaitForAck`. I also think we should avoid the "network may be misconfigured" if any of the indirect pings work. Otherwise, I think we will get lots of spurious warnings when the occasional ping/ack gets dropped.

state.go
slackpad: Haha - I had the bool and took it out because it felt weird to have that extra interface just for that behavior. After sleeping on it, though, I agree with you it would simplify things on the side where things are already complicated, and it'll get rid of the `dial` call outside of the `net` module - I'll change it.

slackpad:
Thanks for the review! I think we should have the property that _any_ UDP ping (direct or indirect) will prevent the TCP warning. Those cases all bail out with a `return` statement when any UDP message is received. It's only if we were about to mark a node suspect where we check if the TCP ping worked.

armon:
@slackpad I may have missed that subtle logic, but glad you thought it through already!

net.go
```diff
@@ -964,3 +988,55 @@ func (m *Memberlist) readUserMsg(bufConn io.Reader, dec *codec.Decoder) error {
 
 	return nil
 }
+
+// sendPingAndWaitForAck makes a TCP connection to the given address, sends
+// a ping, and waits for an ack. All of this is done as a series of blocking
+// operations, given the deadline.
+func (m *Memberlist) sendPingAndWaitForAck(destAddr net.Addr, ping ping, deadline time.Time) (didContact bool, err error) {
```
ryanuber: Just a minor style thing, I'd explicitly return a tuple for clarity in each exit path in this function. It's not functionally different than what you have, but it avoids some assignment and to me is a litter easier to read. Up to you.

net.go
```diff
@@ -964,3 +988,55 @@ func (m *Memberlist) readUserMsg(bufConn io.Reader, dec *codec.Decoder) error {
...
+	if err = dec.Decode(&ack); err != nil {
+		return
+	}
+
+	if ack.SeqNo != ping.SeqNo {
+		err = fmt.Errorf("sequence number from ack (%d) doesn't match ping (%d)", ack.SeqNo, ping.SeqNo)
+		return
+	}
+
+	didContact = true
```
ryanuber: Should this assignment be up higher? It doesn't look like it matters later in the code, but there are some cases where we in fact did contact a node, but return this value false.

state.go
```diff
@@ -265,6 +290,16 @@ func (m *Memberlist) probeNode(node *nodeState) {
 		}
 	}
 
+	// Finally, poll the fallback channel. The timeouts are set such that
+	// the channel will have something or be closed without having to wait
+	// any additional time here.
+	for didContact := range fallbackCh {
+		if didContact {
+			m.logger.Printf("memberlist: Was able to reach %s via TCP but not UDP, network may be misconfigured and not allowing bidirectional UDP", node.Name)
```
ryanuber: We should add a log prefix here, probably `[WARN]`.

net.go
slackpad: Sounds good - I'll make that change.

net.go
slackpad: I wanted it to have to be completely good before we'd consider it a successful ping. Figured a just a positive TCP connect without a proper exchange could be flaky, contact some other service, etc.

ryanuber:
@slackpad this is awesome! Great work. I wonder if we should warn about degraded performance as well, or at least document it some place so that the glacial convergence rate is expected when UDP is not reachable.

slackpad:
Ah good call - I'll update the docs. Thanks for the review!

state.go
```diff
@@ -257,7 +258,31 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
 	}
 
-	// Wait for the acks or timeout
+	// Also make an attempt to contact the node directly over TCP. This
+	// helps prevent confused clients who get isolated from UDP traffic
+	// but can still speak TCP (which also means they can possibly report
+	// misinformation to other nodes via anti-entropy), avoiding flapping in
+	// the cluster.
+	fallbackCh := make(chan bool)
+	if node.PMax >= 3 {
```
armon: We should also check if our own protocol version is at least 3. e.g. both sides must be >= 3

slackpad:
To test this I pulled the new version of the protocol through serf and consul to enable it. With a three node cluster with one leader n1. I logged into n3 and isolated UDP with iptables, one direction at a time:

```
sudo iptables -A OUTPUT -s 172.20.20.10 -p udp -j DROP
sudo iptables -A INPUT -d 172.20.20.10 -p udp -j DROP
```

Without the change, you'd see n1, n2, and n3 get confused and flap, showing nodes dying as pings failed and then coming back as TCP anti-entropy kicks in.

With the change in place, you see the repeating on each node:

```
    2015/06/03 00:34:22 [WARN] memberlist: Was able to reach agent-one via TCP but not UDP, network may be misconfigured and not allowing bidirectional UDP
    2015/06/03 00:34:23 [WARN] memberlist: Was able to reach agent-two via TCP but not UDP, network may be misconfigured and not allowing bidirectional UDP
    2015/06/03 00:34:24 [WARN] memberlist: Was able to reach agent-two via TCP but not UDP, network may be misconfigured and not allowing bidirectional UDP
    2015/06/03 00:34:25 [WARN] memberlist: Was able to reach agent-one via TCP but not UDP, network may be misconfigured and not allowing bidirectional UDP
```

Slightly spammy, but the `consul members` output showed all nodes alive at all times, and the network problem is pretty clear. I was able to restore connectivity by doing:

```
sudo iptables -F 
```

And the messages stopped.

ryanuber:
Sounds right to me! :clap: Looks good from my side.

slackpad:
Hmm - one question from testing - do you think I should check `PMax` or `Pcur` on the other node when deciding to ping it with TCP? I tried joining n3 with `-protocol=2` and the other guys spit TCP pings at it - that feels kind of wrong.

slackpad:
I went ahead and changed it to `pCur` - it seems like the other node needs to be _speaking_ v3 in order to send the ack over TCP, so that's the better check.

armon:
@slackpad I think you want the local node PCur >= 3 and you want the other side PMax >= 3. This way if you only attempt to use TCP mode if you are configured to do so and if the other side understands it.

slackpad:
Ok @armon I reverted that last change and pushed it up. I see it either way - if you are speaking 2 but understand 3 then you can reply to someone over TCP who spoke to you using 3 :-) This'll help this feature get used more widely if consul gets updated but configured to speak 2 because of a mixed environment, the nodes that understand 3 will be able to direct ping each other.

armon:
@slackpad Yeah, it can definitely go either way. I think this way is probably less conservative but gets the feature used in more environments. This LGTM! Merge when you are ready!

armon:
sloth-money

PR #41, opened 2015-06-25
Added Key documentation and godoc image

ErikDubbelboer:
- See https://golang.org/pkg/crypto/aes/#NewCipher for AES docs
- Even though https://golang.org/pkg/crypto/cipher/#NewGCM says it
  works with a 128-bit block cipher it also works with 192 and 256.
  See http://play.golang.org/p/-pXkfFyKE9 as example.


armon:
Thanks!

PR #44, opened 2015-07-17
Add multicast bootstrap for cluster discovery.

oldmantaiter:
- Uses a multicast address and port to bind to on each member to listen for discovery requests
    - Each host listening on the discovery address will respond with its TCP listen address

### DIFF
--- a/config.go
+++ b/config.go
@@ -17,6 +17,20 @@ type Config struct {
 	BindAddr string
 	BindPort int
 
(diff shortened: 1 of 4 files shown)


armon:
We recommend layering this functionality on top like we do in Serf. You can use multicast to discover and issue the join to memberlist. This way they stay decoupled. Multicast is a nightmare and not something we want to embed directly in this library.

PR #47, opened 2015-08-12
Adds more time to check in an attempt to make unit test less flaky.

slackpad:
This timing was a little tight. Lemme know if it still flakes on you.

### DIFF
--- a/state_test.go
+++ b/state_test.go
@@ -117,13 +117,20 @@ func TestMemberList_ProbeNode_FallbackTCP(t *testing.T) {
 
 	var probeTimeMax time.Duration
 	m1 := HostMemberlist(addr1.String(), t, func(c *Config) {
-		c.ProbeTimeout = time.Millisecond
-		c.ProbeInterval = 10 * time.Millisecond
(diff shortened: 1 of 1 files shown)


ryanuber:
Oh computers.... this branch failed on my 3rd attempt on line 173:

```
--- FAIL: TestMemberList_ProbeNode_FallbackTCP-4 (0.02s)
	state_test.go:173: bad seqno 0
```

slackpad:
Hmm I haven't seen that one before. Let me see if I can get that to happen locally.

slackpad:
Looking more carefully at the logic for the random probe it looks like it can give you less than 2 peers if you are unlucky. Changed the logic to make sure we got at least one indirect probe, which should be guaranteed by the algorithm no matter what. I was able to run this in a loop for a few minutes with no failures.

Widened the timing a little more, too, just to be safe.

slackpad:
This one should be good to go if you have a sec.

ryanuber:
LGTM, and passing consistently now. :+1:

PR #48, opened 2015-08-17
discover port on which node binds, if not specified

aybabtme:
When using a configuration with `BindPort=0` so that a random available port be used, I was receiving a bunch of warnings:

```
2015/08/17 18:09:16 [WARN] memberlist: Was able to reach {{hostname}} via TCP but not UDP, network may be misconfigured and not allowing bidirectional UDP
```

Guessing that the problem was the TCP listener was not updating the `BindPort` for the UDP listener, I implemented this fix. It also makes the `NewMemberlistOnOpenPort` function simpler and more likely to succeed.

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -89,6 +89,9 @@ func newMemberlist(conf *Config) (*Memberlist, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Failed to start TCP listener. Err: %s", err)
 	}
+	if conf.BindPort == 0 {
+		conf.BindPort = tcpLn.Addr().(*net.TCPAddr).Port
+	}
 
 	udpAddr := &net.UDPAddr{IP: net.ParseIP(conf.BindAddr), Port: conf.BindPort}
 	udpLn, err := net.ListenUDP("udp", udpAddr)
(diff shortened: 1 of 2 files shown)


memberlist.go
```diff
@@ -126,6 +133,18 @@ func newMemberlist(conf *Config) (*Memberlist, error) {
 	return m, nil
 }
 
+func actualBindPort(l net.Listener) (int, error) {
+	_, portStr, err := net.SplitHostPort(l.Addr().String())
```
armon: You can just cast the `l.Addr()` to `net.TCPAddr` and get the Port directly! Way simpler than having to convert to string, split and parse.

aybabtme: that is a true truth

memberlist.go
```diff
@@ -89,6 +89,9 @@ func newMemberlist(conf *Config) (*Memberlist, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Failed to start TCP listener. Err: %s", err)
 	}
+	if conf.BindPort == 0 {
+		conf.BindPort = tcpLn.Addr().(*net.TCPAddr).Port
```
aybabtme: @armon how does this make you feel? or would you rather

```go
addr, ok := tcpLn.Addr().(*net.TCPAddr)
if !ok { // emit some error }
conf.BindPort = addr.Port
```

armon:
Looks good! Thanks!

PR #49, opened 2015-08-21
Improves accuracy of RTT estimates.

slackpad:
* Plumbs the RX timestamp down from right after the receive call.
 * Moves the TX timestamp after the prep and send call.
 * Gets rid of the ping delegate call for indirect pings (the
   sequencing of the calls here make it hard to improve the
   accuracy without some complicated logic, and this is a degraded
   situation anyway).

### DIFF
--- a/net.go
+++ b/net.go
@@ -281,6 +281,10 @@ func (m *Memberlist) udpListen() {
 			continue
 		}
 
+		// Capture the reception time of the packet as close to the
+		// system calls as possible.
+		lastPacket = time.Now()
+
 		// Check the length
 		if n < 1 {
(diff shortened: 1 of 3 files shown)


slackpad:
This gets me within a few hundred microseconds of what `ping` is reporting, which is a vast improvement. The biggest contribution came from move the sent timestamp to after the actual prep and send. I didn't dig deeper to see what part of it was most expensive, but it's probably the system call.

mitchellh:
Awesome, makes sense LGTM

armon:
Yep, LGTM! Awesome that it was such a huge improvement

PR #50, opened 2015-08-29
Makes the default protocol 2 and lets 3 interoperate with 2.

slackpad:
Not sure how you guys feel about this one, but it should make it easier for people to use Consul 0.6 without having to deal with mismatched protocol versions. If this is accepted, I'll bump Serf and Consul protocol versions back down.

This keeps memberlist speaking v2, but if the other side understands v3, it'll try TCP pings. I added a config option to turn off TCP pings since there was no other way to do that.

### DIFF
--- a/config.go
+++ b/config.go
@@ -85,6 +85,11 @@ type Config struct {
 	ProbeInterval time.Duration
 	ProbeTimeout  time.Duration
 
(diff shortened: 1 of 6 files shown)


armon:
@slackpad It's a bit wonky but makes sense to me.

PR #52, opened 2015-09-16
Log a debug message when a udp ping fails due to timeout

jsternberg:
This can be a helpful debug message when trying to find
misconfigurations of firewalls or EC2 security groups that cause udp
pings to occasionally fail. When a udp ping fails, that might indicate a
problem in configuration and having some debug message about it, with
the information about the node its trying to ping, can be useful in
finding the source of the failure.

This should help with ambiguous issues with flapping as found in
hashicorp/consul#1212.


slackpad:
Indeed - this is useful for finding failed direct pings to particular nodes. Thanks!

Issue #53, opened 2015-09-17
Metrics counter for "alive" isn't accurate


jsternberg:
The metrics counter for when an alive message is processed isn't accurate. The current logic checks the incarnation of the alive node received and will bail if the incarnation is less than or equal to what is currently known for other nodes, but it will only bail when the incarnation is strictly less than when the alive message received is for yourself. This is so the current node can refute an alive message that has the wrong information, but the logic causes a problem with metrics.

When an alive message is received for yourself it always increases the counter even if nothing was done with the message. When a push/pull request happens with another node, the current node rediscovers itself and then will discard the message later when it sees it isn't new. But by the time it discards the message, the node has already reported that it saw an alive message. Since a push/pull happens roughly every 30 seconds, that means you see ~2 alive messages for a node constantly. This implies that there is some activity going on in the system even though there isn't.

I think the increase counter metric should only be sent when an alive message is going to be broadcast rather than always being sent.

PR #54, opened 2015-09-17
Report the alive metric only when an alive message is broadcast

jsternberg:
This fixes hashicorp/memberlist#53.


slackpad:
Makes sense - thanks!

PR #55, opened 2015-11-03
Add Memberlist.Ping() method

jsullivan3:
To enable applications to perform a diagnostic ping of nodes in the data center, the Memberlist.probeNode() method's 'ping' functionality is broken out into a separate Memberlist.pingNode() method, and a new Memberlist.Ping() method allows a consumer to ping other nodes and, if a response is received, record the response time.

### DIFF
--- a/state.go
+++ b/state.go
@@ -48,6 +48,16 @@ type ackHandler struct {
 	timer   *time.Timer
 }
 
+// NoPingResponseError is used to indicate a 'ping' packet was
+// successfully issued but no response was received
+type NoPingResponseError struct {
+	node string
+}
+
+func (f NoPingResponseError) Error() string {
+	return fmt.Sprintf("No response from node %s", f.node)
+}
+
 // Schedule is used to ensure the Tick is performed periodically. This
 // function is safe to call multiple times. If the memberlist is already
 // scheduled, then it won't do anything.
@@ -217,11 +227,18 @@ func (m *Memberlist) probeNode(node *nodeState) {
 
 	// Send a ping to the node.
 	deadline := time.Now().Add(m.config.ProbeInterval)
-	if success, _, err := m.pingNode(&ping, node, ackCh); err != nil {
-		m.logger.Printf("[ERR] memberlist: Could not ping node: %s", err)
+	if rtt, v, err := m.pingNode(&ping, node.Addr, node.Port, ackCh); err == nil {
+		if m.config.Ping != nil {
+			m.config.Ping.NotifyPingComplete(&node.Node, *rtt, v.Payload)
+		}
 		return
-	} else if success {
-		return
+	} else {
+		switch err.(type) {
+		case NoPingResponseError:
+		default:
+			m.logger.Printf("[ERR] memberlist: Could not ping node: %s", err)
+			return
+		}
 	}
 
 	// Get some random live nodes.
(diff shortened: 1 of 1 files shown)


state.go
```diff
@@ -317,6 +292,66 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	m.suspectNode(&s)
 }
 
+// Ping initiates a ping to the node with the specified name.
+func (m *Memberlist) Ping(node string) (bool, time.Duration, error) {
```
ryanbreen: I think it would be cleaner if Ping and pingNode returned duration and error.  `error != nil` should mean the same thing as `success == false`, so there would be no need for the success return value.

jsullivan3: My intent was for `error != nil` to indicate an error transmitting the ping - for example, the node name not being present in the list of members.  The `success` bool indicates whether or not a response was received within `m.config.ProbeTimeout`.  I think keeping both the bool and the error string is valuable since they differentiate between a fatal error that would prevent the ping from being sent, and a potential transient error that prevents the target from responding.  Perhaps renaming `success` to `responseReceived` would clarify the meaning.

One use case for the keeping the bool and error string return values can be seen in the Consul modifications that invoke Memberlist.Ping().  (See https://github.com/jsullivan3/consul/pull/2 for details.)  This patch will continue issuing ping packets to the target node if no response is received, but will not attempt another ping if `error != nil`.

ryanbreen: Yep, that makes sense.  I was thinking you could return a specific error in the retry case and check for that error in the caller to determine whether it's transient or not.  Renaming the return value I think would be helpful in understanding the intent.

jsullivan3: Sounds like the best way to perform this type of error checking is to define a new error type that describes the case in which a ping was successfully transmitted but no response was received.  I'll work on this change and add it to the pull request once it's complete.  Thanks for the help!

state.go
```diff
@@ -317,6 +309,54 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	m.suspectNode(&s)
 }
 
+// Ping initiates a ping to the node with the specified name.
+func (m *Memberlist) Ping(node string, addr net.IP, port uint16) (*time.Duration, error) {
```
slackpad: It feels a little weird to have a pointer return for the RTT, can you change that to a regular value and just return 0 in the error cases?

state.go
```diff
@@ -317,6 +309,54 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
+
+	// Wait for response or round-trip-time.
+	select {
+	case v := <-ackCh:
+		if v.Complete == true {
+			rtt := v.Timestamp.Sub(sent)
+			return &rtt, &v, nil
+		}
+
+		// As an edge case, if we get a timeout, we need to re-enqueue it
```
slackpad: Hmm - this is referring to a `select()` that's actually up in `probeNode()`. This is kind of tricky given how things are layered here and how the `probeNode()` runs other stuff in parallel with this one. It wouldn't be good to always add it back in at this level.

I'm kind of thinking that given how this code isn't very complex we should just leave `probeNode()` alone and not try to share code. Adding a new `Ping()` method will only be a dozen lines of code, and would keep things simpler overall.

slackpad:
Hi @jsullivan3 - what do you think about https://github.com/hashicorp/memberlist/pull/55#discussion-diff-46748632? I think just add a standalone `Ping()` function and things will be simpler - there's not a whole lot of shared code other than the RTT calculation, but that's pretty simple.

jsullivan3:
I can get behind simplifying the change so that Ping() is a standalone function.  I was aiming for commonality, but there's certainly a case to be made for simplicity.  I'll test and push this change within the next couple of days.  Thanks!

jsullivan3:
Hi @slackpad - I implemented the changes you suggested and updated https://github.com/hashicorp/consul/pull/1388 to reflect the change in RTT return type.  Thank you for your review comments - please let me know if you have any additional suggestions.

slackpad:
HI @jsullivan3 I made a few minor tweaks and merged this under https://github.com/hashicorp/memberlist/pull/64. Thanks!

Issue #39, opened 2015-06-13
Memberlist logger should be exported


cousine:
Hi there, 

I think the logger should be exported instead of the LogOutput in the configuration; I'm using a custom logger with a custom format however I must hack a custom io.Writer to have memberlist use my custom logger.

ryanuber:
Hey @cousine, seems reasonable to use a custom logger. Instead of exporting it directly in the struct, maybe we could do like a `CreateLogger(*Config, *log.Logger)` analogous to the `Create` method or something to that effect.

cousine:
Yeah that would be very helpful :+1: but I'd call it `CreateWithLogger` tbh :smile:

JensRantil:
See also https://github.com/hashicorp/memberlist/issues/76.

PR #56, opened 2015-11-10
Adding in the ability to use a custom logger in place of the default one

StabbyCutyou:
This is to address #39 

I figured since you already had the config abstraction, the best thing to do would be to layer this into the existing logoutput options, but to leave those in to maintain compatibility.

### DIFF
--- a/config.go
+++ b/config.go
@@ -135,11 +135,14 @@ type Config struct {
 	Merge                   MergeDelegate
 
 	// LogOutput is the writer where logs should be sent. If this is not
-	// set, logging will go to stderr by default.
+	// set, logging will go to stderr by default. You cannot specify both LogOutput
+	// and Logger at the same time.
 	LogOutput io.Writer
-	// Logger is a custom logger which you provide. If Logger is set, it will use this for
-	// the internal logger. If this is not set, it will fall back to the behavior for
-	// LogOutput is not set, logging will go to stderr by default.
+
+	// Logger is a custom logger which you provide. If Logger is set, it will use
+	// this for the internal logger. If Logger is not set, it will fall back to the
+	// behavior for using LogOutput. You cannot specify both LogOutput and Logger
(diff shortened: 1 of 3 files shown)


ErikDubbelboer:
If this is added I vote to remove `config.LogOutput`. Having both is a little bit confusing and `LogOutput` is not really needed if you have `Logger`. It would break API compatibility though.

StabbyCutyou:
Thats why I left it in - since theres no fantastic way to set a dependency version in a standard way, I didn't want to break the API compatibility. I figured it was safer to just do it this way.

config.go
```diff
@@ -136,6 +137,10 @@ type Config struct {
 	// LogOutput is the writer where logs should be sent. If this is not
 	// set, logging will go to stderr by default.
 	LogOutput io.Writer
+	// Logger is a custom logger which you provide. If Logger is set, it will use this for
+	// the internal logger. If this is not set, it will fall back to the behavior for
+	// LogOutput is not set, logging will go to stderr by default.
+	Logger *log.Logger
```
ryanuber: Can we add a line break between this and the field above? The description could also be cleaned up to be more clear about the LogOutput writer being used when Logger is nil. We should also add a similar note to the LogOutput fie.

ryanuber:
This seems mostly fine to me. It would make sense to return an error if both LogOutput and Logger are non-nil rather than to silently pick one. If we can also add a test case for this, that would be great!

StabbyCutyou:
Happy to add test cases - Didn't wanna put that time in yet in case the PR
was rejected, but can definitely add those in.

If people have suggestions for better comments, I'm happy to incorporate
that as well.

StabbyCutyou:
Changed the comments around a bit, but can edit them some more if you like.

Added the test case for having both set return an error.

slackpad:
This LGTM - thanks!

PR #57, opened 2015-11-11
Only call the delegate with userBuf if it is not nil

jacksontj:
Without this check if a pushPullMsg is sent with a UserStateLen of 0, userBuf is nill and passed to the delegates.

Issue over in serf: https://github.com/hashicorp/serf/issues/339


slackpad:
This looks good - thanks!

PR #58, opened 2015-11-13
Add experimental join for return not joining addresses

saromanov:
My main motivation of this PR is returning addresses which is not joined. It could be helpful for serf in ```join``` command where after this command, showing not only number of joining nodes.

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -185,6 +185,37 @@ func (m *Memberlist) Join(existing []string) (int, error) {
 	return numSuccess, retErr
 }
 
+func (m *Memberlist) JoinNew(existing []string) (int, []string, error) {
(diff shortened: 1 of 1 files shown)


slackpad:
Hi @saromanov - usually when doing a `join` a user is just interested in if they found _some_ node to join with, and it's pretty typical to pass in nodes that might not all be there, such as to make bootstrapping easier and avoid careful sequencing when starting things up. For this reason, I'm not sure that it would be a good addition to this or Serf. Did you have a more specific use in mind that I'm missing?

slackpad:
Hi @saromanov I'm going to go ahead and close this one out - I think this doesn't fit the typical use case as described above.

PR #61, opened 2015-12-08
Keeps logger config options intact so we can re-use a config.

slackpad:
The change in https://github.com/hashicorp/memberlist/pull/56 ended up causing a small regression - https://travis-ci.org/hashicorp/consul/builds/95523949. The problem is that the operations at startup leave the config in a state that won't pass the sanity check the second time around. This PR keeps it from altering the config object when creating a logger, so the config can be reused multiple times. This probably affects unit tests and not real uses of memberlist, but it's bad to leave the config in an invalid state, so it's best fixed down here.


ryanuber:
Makes sense, LGTM

PR #62, opened 2015-12-16
Adds source address logging when errors occur.

slackpad:
This helps you figure out where bad messages are coming from.

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -1,0 +1,22 @@
+package memberlist
+
+import (
+	"fmt"
+	"net"
+)
+
+func LogAddress(addr net.Addr) string {
+	if addr == nil {
+		return "from=<unknown address>"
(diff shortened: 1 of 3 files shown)


mitchellh:
If you want to avoid the mock structures I believe you can also use this: https://golang.org/pkg/net/#ResolveIPAddr

The interface is simple enough, though.

mitchellh:
And for `net.Conn` I usually just do this:

```
ln, err := net.Listen("tcp", ":0")
...
net.Dial("tcp", ln.Addr()) or something like that
```

Just to again avoid just mock heavy stuff.

slackpad:
@mitchellh thanks for the tips! similar number of code lines but less lame mock stuff, especially for `Conn`

PR #64, opened 2015-12-19
Adds Ping method to memberlist.

slackpad:
I took https://github.com/hashicorp/memberlist/pull/55 and did just a few tweaks to it.

### DIFF
--- a/state.go
+++ b/state.go
@@ -48,6 +48,16 @@ type ackHandler struct {
 	timer   *time.Timer
 }
 
+// NoPingResponseError is used to indicate a 'ping' packet was
+// successfully issued but no response was received
+type NoPingResponseError struct {
+	node string
+}
+
+func (f NoPingResponseError) Error() string {
+	return fmt.Sprintf("No response from node %s", f.node)
+}
+
 // Schedule is used to ensure the Tick is performed periodically. This
 // function is safe to call multiple times. If the memberlist is already
 // scheduled, then it won't do anything.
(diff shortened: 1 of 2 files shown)


state.go
```diff
@@ -317,6 +325,38 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	m.suspectNode(&s)
 }
 
+// Ping initiates a ping to the node with the specified name.
+func (m *Memberlist) Ping(node string, addr net.IP, port uint16) (time.Duration, error) {
+	// Prepare a ping message and setup an ack handler.
+	ping := ping{SeqNo: m.nextSeqNo(), Node: node}
+	ackCh := make(chan ackMessage, m.config.IndirectChecks+1)
+	m.setAckChannel(ping.SeqNo, ackCh, m.config.ProbeInterval)
```
ryanuber: What happens if we have two concurrent calls to `Ping()`? Since this method is public that might cause some confusing results. Maybe we should wrap this in a mutex, or at least document that it is not allowed.

ryanuber: Oh I see - the SeqNo makes this nice. NVM!

state.go
```diff
@@ -317,6 +325,38 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
+	// Mark the sent time here, which should be after any pre-processing and
+	// system calls to do the actual send. This probably under-reports a bit,
+	// but it's the best we can do.
+	sent := time.Now()
+
+	// Wait for response or timeout.
+	select {
+	case v := <-ackCh:
+		if v.Complete == true {
+			return v.Timestamp.Sub(sent), nil
```
ryanuber: Does this mean that the returned duration is one-way? (since we are doing math on the timestamp from the payload instead of current time). If so maybe we can slap an extra comment on the function so it will show up in GoDoc.

ryanuber:
@slackpad left an open question but otherwise this all LGTM

state.go
slackpad: It's actually the two-way timestamp, though the plumbing of that is pretty convoluted. The timestamp comes from here https://github.com/hashicorp/memberlist/blob/master/net.go#L295 and gets passed down through function calls and then stuck in the channel for this to read out. This tries to get the timestamp up as close to the IO as possible.

PR #65, opened 2015-12-19
Changes ping interface to take a regular net.Addr.

slackpad:


ryanuber:
Yeah this is better - LGTM

PR #68, opened 2016-02-25
Centralize private IP network handling into a single place

sean-:
Consul and memberlist both have copies of this.  Consolidate this to memberlist for now.


slackpad:
LGTM

Issue #63, opened 2015-12-19
IPv6 parsing in resolveAddr doesn't handle the case where the port is missing


slackpad:
This came out of https://github.com/hashicorp/serf/issues/286 but the problem is down in memberlist. Behavior may have varied across Go versions, but with 1.5.2 if you don't supply a port number then it will try to lookup the address via DNS. Adding the port number does the right thing.

We should fix the support for IPV6 addresses without port numbers in`resolveAddr()`.

meyskens:
I was looking into https://github.com/moby/moby/issues/33398 which seems to be an issue in memberlist rather than moby. Seeing the PR is this fixed or not?

slackpad:
Hi @meyskens this should have been closed with #69. Thanks for adding the additional test!

PR #69, opened 2016-03-05
Fix resolution of IPv6 addresses

orivej:
Previously `resolveAddr("[IPv6-addr]")` passed `[IPv6-addr]` to `net.ParseIP` and failed with `invalid domain name`.

`hasPort` utility comes from `net/http/client.go`.

This would fix `consul join "[IPv6-addr]"` and #63.


slackpad:
Thanks!

PR #70, opened 2016-03-05
Report all errors when join fails

orivej:
This would report all join errors.

Previous behavior was confusing e.g. in consul when consul was concealing important errors and displaying just the last irrelevant one.

The new info log message might seem redundant, yet it is not useless because it always reports attempted IP, and it is greppable.

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -15,11 +15,13 @@ multiple routes.
 package memberlist
 
 import (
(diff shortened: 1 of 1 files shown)


slackpad:
Thanks for this - I switched it to use go-multierror so the caller can format the list of errors how they like, but they aren't required to.

PR #71, opened 2016-03-05
Fix goroutine leak when UDP pings fail but TCP pings succeed

jacksontj:
In the case that the TCP ping succeeds and not the UDP ping, we call return before ever reading the `fallbackCh`. Since the `fallbackCh` was an unbuffered channel it would block that goroutine indefinitely. I see this occasionally in production with traces (from pprof) that look like:
```
goroutine 388015741 [chan send, 1120 minutes]:
github.com/hashicorp/memberlist.(*Memberlist).probeNode.func1(0xc824277560, 0xc8275d27e0, 0xc848aff3b0, 0xc83e8f65c0,      0xece6bede8, 0xc835701734, 0x1402440)
       src/github.com/hashicorp/memberlist/state.go:294 +0x245
created by github.com/hashicorp/memberlist.(*Memberlist).probeNode
        src/go/pkg/src/github.com/hashicorp/memberlist/state.go:296 +0xea7
```

A simple solution is to simply make the `fallbackCh` buffered, since it will only ever send a single item. This way if we return without reading the `fallbackCh` the goroutine will complete because it wrote to the channel-- and since no one will reference the channel it will be properly GCd.

### DIFF
--- a/state.go
+++ b/state.go
@@ -282,7 +282,7 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	// member who understands version 3 of the protocol, regardless of
 	// which protocol version we are speaking. That's why we've included a
 	// config option to turn this off if desired.
-	fallbackCh := make(chan bool)
+	fallbackCh := make(chan bool, 1)
 	if (!m.config.DisableTcpPings) && (node.PMax >= 3) {
 		destAddr := &net.TCPAddr{IP: node.Addr, Port: int(node.Port)}
 		go func() {


jacksontj:
@slackpad Not sure if you are the right person to ping on this, but you did respond to my last patch. This is a fairly important fix, as UDP network problems can cause fairly severe goroutine leaks.

slackpad:
Hi @jacksontj thanks for the fix and the patch - this looks good!

PR #77, opened 2016-03-29
Adds a TCP-first DNS lookup for joins.

### LINKED ISSUE #74: Perform on-join DNS requests over TCP
If we make the request via TCP we can more easily get a larger initial join list from the DNS server. Right now the default UDP client will only get a short list of responses (~3). If we pull SRV records we can get the port information as well.

slackpad:
This fixes #74.

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -20,10 +20,12 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	"github.com/miekg/dns"
 )
 
 type Memberlist struct {
@@ -168,11 +170,10 @@ func Create(conf *Config) (*Memberlist, error) {
 // none could be reached. If an error is returned, the node did not successfully
 // join the cluster.
 func (m *Memberlist) Join(existing []string) (int, error) {
-	// Attempt to join any of them
 	numSuccess := 0
 	var errs error
 	for _, exist := range existing {
-		addrs, port, err := m.resolveAddr(exist)
+		addrs, err := m.resolveAddr(exist)
 		if err != nil {
 			err = fmt.Errorf("Failed to resolve %s: %v", exist, err)
 			errs = multierror.Append(errs, err)
@@ -181,8 +182,8 @@ func (m *Memberlist) Join(existing []string) (int, error) {
 		}
 
 		for _, addr := range addrs {
-			if err := m.pushPullNode(addr, port, true); err != nil {
-				err = fmt.Errorf("Failed to join %s: %v", net.IP(addr), err)
+			if err := m.pushPullNode(addr.ip, addr.port, true); err != nil {
+				err = fmt.Errorf("Failed to join %s: %v", net.IP(addr.ip), err)
 				errs = multierror.Append(errs, err)
 				m.logger.Printf("[DEBUG] memberlist: %v", err)
 				continue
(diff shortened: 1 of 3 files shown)


memberlist.go
```diff
@@ -191,50 +191,111 @@ func (m *Memberlist) Join(existing []string) (int, error) {
...
+	if len(cc.Servers) > 0 {
+		server := cc.Servers[0]
+		if !hasPort(server) {
+			server = net.JoinHostPort(server, cc.Port)
+		}
+
+		c := new(dns.Client)
+		c.Net = "tcp"
+		m := new(dns.Msg)
+		m.SetQuestion(host, dns.TypeANY)
```
sean-: This comment is purely informational as there's nothing better to be done, but it's sad that this was brought up as a problem in '98 and it is still a problem (despite repeated attempts to fix over the years).  Item 4 in the following has additional detail.

http://tools.ietf.org/html/draft-sajitha-dnsext-qtype-addr-00

TL;DR: `ANY` is as good as it gets and works for the `A` and `AAAA` QTYPES, but it won't give the DNS server the opportunity to resolve `CNAME` records and return the correct information in the `ADDITIONAL` section.  Hold this thought for a second.

`T_ANY` also has the server xmit all records to the client, which shouldn't be common, and as you said, this is reasonably rare.

memberlist.go
```diff
@@ -191,50 +191,111 @@ func (m *Memberlist) Join(existing []string) (int, error) {
...
+		}
+
+		var ips []ipPort
+		for _, r := range in.Answer {
+			switch rr := r.(type) {
+			case (*dns.A):
+				ips = append(ips, ipPort{rr.A, defaultPort})
+			case (*dns.AAAA):
+				ips = append(ips, ipPort{rr.AAAA, defaultPort})
+			}
```
sean-: It might be worth handling `CNAME` or emitting a log message if there's a logger in scope about CNAME not being supported that response type comes back.  Intra-organization DNS shouldn't be using CNAMEs, but it's not unheard of.  The common case when talking to Consul doesn't have this problem, however.

memberlist.go
```diff
@@ -191,50 +191,111 @@ func (m *Memberlist) Join(existing []string) (int, error) {
...
-
 	if numSuccess > 0 {
 		errs = nil
 	}
 	return numSuccess, errs
 }
 
+// ipPort holds information about a node we want to try to join.
+type ipPort struct {
+	ip   []byte
```
sean-: Is there any reason the type shouldn't be `net.IP`?  `[]byte` is the backing store for `net.IP` so it should be equivalent.

sean-:
LGTM

memberlist.go
slackpad: Good catch - I'll do that.

slackpad:
@sean- thanks for the review. I didn't realize you were on it and force-pushed a change to add the “.” at the end if not there, and to not try local names without a “.” with the TCP resolver

memberlist.go
slackpad: I'll add a debug log about that (i could just emit a log for any other type of answer that i'm ignoring)

sean-: That would be way too spammy (think of zones secured with `DNSSEC`).  IMO, the only `QTYPE` that's worth emitting a warning on is `CNAME` and everything else can be safely ignored.

sean-:
Heh, no worries. GitHub is rather good about preserving comments across page reloads (props to whoever spent effort on making that happen at @github).

memberlist.go
slackpad: copy - will reduce to just `CNAME`

PR #82, opened 2016-05-27
Adds a "dogpile" feature where a large initial suspect counter is accelerated by confirmations from peers.

slackpad:
The dogpile part ended up being big enough that I'll submit this for review now and add the other parts in separate PRs branched from this one.

### DIFF
--- a/state.go
+++ b/state.go
@@ -694,9 +694,6 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 		}
 	}
 
-	// Clear out any suspicion timer that may be in effect.
-	delete(m.nodeTimers, a.Node)
-
 	// Check if we've never seen this node before, and if not, then
 	// store this node in our node map.
 	if !ok {
@@ -756,6 +753,9 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 	if a.Incarnation < state.Incarnation && isLocalNode {
 		return
 	}
+
+	// Clear out any suspicion timer that may be in effect.
+	delete(m.nodeTimers, a.Node)
 
 	// Store the old state and meta data
 	oldState := state.State
--- a/state.go
+++ b/state.go
@@ -854,9 +854,9 @@ func (m *Memberlist) suspectNode(s *suspect) {
 		return
 	}
 
-	// See if there's a suspicion timer we can corroborate
+	// See if there's a suspicion timer we can confirm
 	if timer, ok := m.nodeTimers[s.Node]; ok && s.From != m.config.Name {
-		timer.Corroborate(s.From)
+		timer.Confirm(s.From)
 		return
 	}
 
--- a/state.go
+++ b/state.go
@@ -910,9 +910,10 @@ func (m *Memberlist) suspectNode(s *suspect) {
 		k = 1
 	}
 	n := m.estNumNodes()
+	peersAvailable := n-2 >= k // take ourselves out plus the node being probed
 	timeout := suspicionTimeout(m.config.SuspicionMult, n, m.config.ProbeInterval)
 	bound := time.Duration(m.config.SuspicionMaxTimeoutMult) * timeout
-	if n <= k {
+	if !peersAvailable {
 		bound = timeout
 	}
 	f := func(numConfirmations int32) {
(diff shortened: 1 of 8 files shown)


state.go
```diff
@@ -694,6 +694,9 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 		}
 	}
 
+	// Clear out any suspicion timer that may be in effect.
+	delete(m.nodeTimers, a.Node)
```
armon: We should probably move this below all the checks of incarnation number to avoid an old packet from causing the timer to get reset

suspicion.go
```diff
@@ -0,0 +1,89 @@
...
+	// a way the achieves the overall time we'd like.
+	start time.Time
+
+	// timer is the underlying timer that implements the timeout.
+	timer *time.Timer
+
+	// n is the number of independent confirmations we've seen. This must
+	// be updated using atomic instructions to prevent contention with the
+	// timer callback.
+	n int32
```
armon: Move this to the top of the struct, otherwise you can get weird alignment issues for atomic loads

suspicion.go
```diff
@@ -0,0 +1,89 @@
...
+// suspicion manages the suspect timer for a node and provides an interface
+// to accelerate the timeout as we get more independent confirmations that
+// a node is suspect.
+type suspicion struct {
+	// k is the number of independent confirmations we'd like to see in
+	// order to drive the timer to its minimum value.
+	k int32
+
+	// min is the minimum timer value in seconds.
+	min float64
```
armon: Why not keep these as time.Duration?

suspicion.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Only allow one confirmation from each possible peer.
+	if _, ok := s.confirmations[from]; ok {
+		return false
+	}
+	s.confirmations[from] = struct{}{}
+
+	// Compute the new timeout given the current number of confirmations.
+	n := float64(atomic.AddInt32(&s.n, 1))
+	timeout := math.Max(s.min, s.max-(s.max-s.min)*math.Log(n+1.0)/math.Log(float64(s.k)+1.0))
```
armon: We can make this easier to read by unfolding the min/max. e.g. `if timeout < s.min { timeout = s.min }`

suspicion.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Reset the timer. We have to take into account the amount of time that
+	// has passed so far, so we get the right overall timeout.
+	remaining := math.Max(0.0, s.start.Sub(time.Now()).Seconds()+timeout)
```
armon: Is the remaining time not (timeout - elapsed) = (timeout - (time.Now() - start)?
If remaining goes <= 0 we can just fire the timer / callback immediately.

suspicion_test.go
```diff
@@ -0,0 +1,135 @@
...
+				pair{"me", false},
+				pair{"foo", true},
+				pair{"bar", true},
+				pair{"baz", true},
+				pair{"zoo", false},
+			},
+			min,
+		},
+	}
+	for i, c := range cases {
```
armon: Might be easier to test if the struct split off the method of computing the timeout from the actual timer handling, then you can just test the logic instead of the timer interactions

state.go
```diff
@@ -894,26 +908,40 @@ func (m *Memberlist) suspectNode(s *suspect) {
...
 
-	// Setup a timeout for this
-	timeout := suspicionTimeout(m.config.SuspicionMult, m.estNumNodes(), m.config.ProbeInterval)
-	time.AfterFunc(timeout, func() {
+	// Setup a suspicion timer. Given that we don't have any known phase
+	// relationship with our peers, we set up k such that we hit the nominal
+	// timeout two probe intervals short of what we expect given the suspicion
+	// multiplier. If there aren't enough peers then we just use the nominal
+	// timer, since we know this algorithm no longer makes sense.
+	k := m.config.SuspicionMult - 2
```
armon: I feel like we should move some of this logic into a helper function as its pretty tricky

suspicion.go
slackpad: All the math is on them as floats but `time.Duration` is probably more readable / safer for later maintenance so I'll change it.

suspicion.go
slackpad: Yeah - since I was working with raw seconds the backwards order gave me the right sign. I'll rework this to get into `time.Duration` earlier which will make this less confusing.

state.go
```diff
@@ -894,26 +908,40 @@ func (m *Memberlist) suspectNode(s *suspect) {
...
+	if k < 1 {
+		k = 1
+	}
+	n := m.estNumNodes()
+	peersAvailable := n-2 >= k // take ourselves out plus the node being probed
```
slackpad: Lower `k` instead

PR #83, opened 2016-06-03
Adds a "self-awareness" feature that slows down the failure detector if a node is unhealthy.

slackpad:
This uses several signals to slow down the failure detector if a node might not be meeting the soft real-time requirements of the protocol. Things that slow the detector down:

1. Failed probe if no peers support nacks
2. Missing nacks if the peers should support it
3. Having to refute a suspect or dead message

Things that speed the detector back up:

1. Successful probe of another node

We added nacks as part of this change, so a node can get feedback from an indirect probe, even if the node being probed is actually dead.

### DIFF
--- a/config.go
+++ b/config.go
@@ -205,7 +205,7 @@ func DefaultLANConfig() *Config {
 		ProbeTimeout:            500 * time.Millisecond, // Reasonable RTT time for LAN
 		ProbeInterval:           1 * time.Second,        // Failure check every second
 		DisableTcpPings:         false,                  // TCP pings are safe, even with mixed versions
-		AwarenessMaxMultiplier:  7,                      // Probe interval backs off to (7 + 1) seconds
+		AwarenessMaxMultiplier:  8,                      // Probe interval backs off to 8 seconds
 
 		GossipNodes:    3,                      // Gossip to 3 nodes
 		GossipInterval: 200 * time.Millisecond, // Gossip more rapidly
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -200,6 +205,7 @@ func DefaultLANConfig() *Config {
 		ProbeTimeout:            500 * time.Millisecond, // Reasonable RTT time for LAN
 		ProbeInterval:           1 * time.Second,        // Failure check every second
 		DisableTcpPings:         false,                  // TCP pings are safe, even with mixed versions
+		AwarenessMaxMultiplier:  7,                      // Probe interval backs off to (7 + 1) seconds
```
armon: Is the maximum 7 seconds (e.g. its multiplied by 7) or its scaled by 7 (max is 8). Don't mean to be pedantic but its just unclear to me

net.go
```diff
@@ -440,29 +457,49 @@ func (m *Memberlist) handleIndirectPing(buf []byte, from net.Addr) {
...
 	}
 
-	// Send a ping to the correct host
+	// Send a ping to the correct host.
 	localSeqNo := m.nextSeqNo()
 	ping := ping{SeqNo: localSeqNo, Node: ind.Node}
 	destAddr := &net.UDPAddr{IP: ind.Target, Port: int(ind.Port)}
 
 	// Setup a response handler to relay the ack
+	cancelCh := make(chan struct{}, 1)
```
armon: Don't need to buffer it, since we don't actually ever write to it

config.go
slackpad: 8 -> we always add 1 when for when the health score is 0

slackpad: I'll make the math work out so we can set this to 8 :-)

PR #84, opened 2016-06-03
Adds a hint during a probe that a node is thought to be degraded.

slackpad:
This allows the node being probed to immediately start gossiping a
refutation, rather than wait for an event to come in via gossip that
it can react to.

### DIFF
--- a/net.go
+++ b/net.go
@@ -82,6 +82,12 @@ type ping struct {
 	// the intended recipient. This is to protect again an agent
 	// restart with a new name.
 	Node string
+
+	// YouShouldRefute is a courtesy flag that's sent during a probe to
+	// tell the node being probed that it's currently in the suspect or dead
+	// state from the perspective of the prober so it can start right away
+	// trying to refute that.
+	YouShouldRefute bool
 }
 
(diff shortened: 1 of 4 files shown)


net.go
```diff
@@ -82,6 +82,12 @@ type ping struct {
 	// the intended recipient. This is to protect again an agent
 	// restart with a new name.
 	Node string
+
+	// YouShouldRefute is a courtesy flag that's sent during a probe to
+	// tell the node being probed that it's currently in the suspect or dead
+	// state from the perspective of the prober so it can start right away
+	// trying to refute that.
+	YouShouldRefute bool
```
armon: Can we call this `SuspectFailed` instead?

armon: Also I wonder if it makes sense to include the incarnation number of the suspect. e.g. if I "suspect" at 7 but the alive for 10 is already out, then its fine and can be no-oped.

armon: A cleaner way to do this might be to just staple a suspect message and do a compound message.

slackpad:
Going to go with a suspect message, which is was simpler.

PR #87, opened 2016-06-29
Expose a method to validate a gossip encryption key before use.

rboyer:
This is used in the fix for https://github.com/hashicorp/consul/issues/1299


rboyer:
Is there anything I can do to move this along?

slackpad:
Thanks! Sorry for the delay on this.

PR #89, opened 2016-09-09
Add configuration option for size of handoff channel

tyler:
This allows the end user to modify the size of the channel which is used in handling incoming messages. I've been using Serf for a project that sometimes has large spikes in messages. If those spikes happen to be paired with a small delay in handling the messages, it results in a deluge of "UDP handler queue full, dropping message".

### DIFF
--- a/config.go
+++ b/config.go
@@ -238,7 +238,6 @@ func DefaultWANConfig() *Config {
 	conf.ProbeInterval = 5 * time.Second
 	conf.GossipNodes = 4 // Gossip less frequently, but to an additional node
 	conf.GossipInterval = 500 * time.Millisecond
-	conf.HandoffQueueDepth = 1024
 	return conf
 }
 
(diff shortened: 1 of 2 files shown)


tyler:
Hey folks. Is there anything I can do to help move this along?

config.go
```diff
@@ -231,6 +238,7 @@ func DefaultWANConfig() *Config {
 	conf.ProbeInterval = 5 * time.Second
 	conf.GossipNodes = 4 // Gossip less frequently, but to an additional node
 	conf.GossipInterval = 500 * time.Millisecond
+	conf.HandoffQueueDepth = 1024
```
slackpad: You can remove this and the one on line 258 since these both inherit from the LAN config, that way these only specify things that change for WAN and local.

tyler:
Good point. I hadn't noticed that they were initialized from DefaultLANConfig. Thanks!

slackpad:
Thanks!

PR #91, opened 2016-09-23
Expect addr like ::1 in hasPort util

sakateka:
resolv.conf support v6 addr without '[]', hasPort should handle this case.

### DIFF
--- a/util.go
+++ b/util.go
@@ -327,10 +327,19 @@ func isLoopbackIP(ip_str string) bool {
 	return loopbackBlock.Contains(ip)
 }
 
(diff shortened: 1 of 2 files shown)


sakateka:
@slackpad Please take a look.

sakateka:
Hi, what can be done to make this PR have merged?

sakateka:
@ryanuber,  please take a look.

sakateka:
Hello, write any reply.

slackpad:
@sakateka sorry for the delay on that one!

PR #94, opened 2016-11-15
Add configuration option for UDP packet size

tyler:
Hello there!

This PR adds a configuration option, `UDPBufferSize`, that sets the expected size of UDP packets. 1400 is a completely reasonable default, but depending on the network that you're running Memberlist on, you may be able to safely set it higher. In my case, this is useful because the MTU is set to 9000, which ends up increasing throughput by about 6x. 

Let me know if there's anything I need to do to make this more acceptable. :)

Thanks.

### DIFF
--- a/config.go
+++ b/config.go
@@ -184,6 +184,12 @@ type Config struct {
 	// size of this determines the size of the queue which Memberlist will keep
 	// while UDP messages are handled.
 	HandoffQueueDepth int
+
+	// Maximum number of bytes that memberlist expects UDP messages to be. A safe
+	// value for this is typically 1400 bytes (which is the default.) However,
+	// depending on your network's MTU (Maximum Transmission Unit) you may be able
+	// to increase this.
+	UDPBufferSize int
 }
 
 // DefaultLANConfig returns a sane set of configurations for Memberlist.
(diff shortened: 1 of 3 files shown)


slackpad:
LGTM - thanks!

slackpad:
@tyler merged this because it looks good, though I realized after that we should probably have the RX side be based on the send side:

https://github.com/hashicorp/memberlist/blob/master/memberlist.go#L110-L111

This will be too small of an RX buffer if you set a large MTU.

tyler:
@slackpad Good point.

It looks like that is set to 2MB based on the constant here: https://github.com/hashicorp/memberlist/blob/5d2e9d64179cfb6d18f700e5bf55eb99a05de949/net.go#L72

Do you know how the value of 2MB was originally derived? It seemed sufficient for the MTU sizes that I've observed, but I had not considered *really* large MTUs. :)

slackpad:
Hi @tyler not sure where the size came from, but it looks like it has been that way for a long time. I realized that I misread this as `2 * 1024` not `2 * 1024 * 1024`. This is probably ok as-is unless people start to experience issues, as it's pretty huge. We could put a separate config in if we ever need it - just something to be aware of.

Intar:
Hello @tyler
Could we use value from config here?
https://github.com/hashicorp/memberlist/blob/master/net_transport.go#L269
We see a lot of allocations in this place and it seems like unnecessary.

tyler:
Yeah, if I’m reading that code correctly, it would make sense to have that set to the MTU.

Intar:
Created [PR](https://github.com/hashicorp/memberlist/pull/188) for use this option in that place too.

PR #95, opened 2016-12-01
Add a checksum for UDP messages

kyhavlov:
Adds a checksum to detect UDP corruption as described in hashicorp/consul#2519.

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -495,18 +495,8 @@ func (m *Memberlist) SendTo(to net.Addr, msg []byte) error {
 	buf[0] = byte(userMsg)
 	buf = append(buf, msg...)
 
-	// Lookup node from address
-	toAddr := strings.Split(to.String(), ":")[0]
-	m.nodeLock.RLock()
-	nodeState, ok := m.nodeMap[toAddr]
-	m.nodeLock.RUnlock()
-	var node *Node
-	if ok {
-		node = &nodeState.Node
-	}
-
 	// Send the message
-	return m.rawSendMsgUDP(to, node, buf)
+	return m.rawSendMsgUDP(to, nil, buf)
 }
 
 // SendToUDP is used to directly send a message to another node, without
--- a/memberlist.go
+++ b/memberlist.go
@@ -495,8 +495,18 @@ func (m *Memberlist) SendTo(to net.Addr, msg []byte) error {
 	buf[0] = byte(userMsg)
 	buf = append(buf, msg...)
 
+	// Lookup node from address
+	toAddr := strings.Split(to.String(), ":")[0]
+	m.nodeLock.RLock()
+	nodeState, ok := m.nodeMap[toAddr]
+	m.nodeLock.RUnlock()
+	var node *Node
+	if ok {
+		node = &nodeState.Node
+	}
+
 	// Send the message
-	return m.rawSendMsgUDP(to, buf)
+	return m.rawSendMsgUDP(to, node, buf)
 }
 
 // SendToUDP is used to directly send a message to another node, without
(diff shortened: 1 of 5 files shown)


memberlist.go
```diff
@@ -495,8 +495,18 @@ func (m *Memberlist) SendTo(to net.Addr, msg []byte) error {
 	buf[0] = byte(userMsg)
 	buf = append(buf, msg...)
 
+	// Lookup node from address
+	toAddr := strings.Split(to.String(), ":")[0]
```
slackpad: This'll crash if the split doesn't work - let's use https://golang.org/pkg/net/#SplitHostPort and check the error return from that.

memberlist_test.go
```diff
@@ -367,6 +367,8 @@ func (h dnsHandler) ServeDNS(w dns.ResponseWriter, r *dns.Msg) {
 }
 
 func TestMemberList_ResolveAddr_TCP_First(t *testing.T) {
+	t.Skip()
```
slackpad: This is a leftover I think.

net.go
```diff
@@ -599,9 +613,19 @@ func (m *Memberlist) sendMsg(to net.Addr, msg []byte) error {
 	}
 	extra := m.getBroadcasts(compoundOverhead, bytesAvail)
 
+	// Lookup node from address
+	toAddr := strings.Split(to.String(), ":")[0]
```
slackpad: Same for this one as above.

net.go
```diff
@@ -338,8 +341,19 @@ func (m *Memberlist) ingestPacket(buf []byte, from net.Addr, timestamp time.Time
 		buf = plain
 	}
 
-	// Handle the command
-	m.handleCommand(buf, from, timestamp)
+	// See if there's a checksum included to verify the contents of the message
+	msgType := messageType(buf[0])
```
slackpad: Need a size check to make sure buf isn't empty, otherwise this can panic.

slackpad: You actually need to check for at least length 5.

net.go
```diff
@@ -338,8 +341,19 @@ func (m *Memberlist) ingestPacket(buf []byte, from net.Addr, timestamp time.Time
...
+	if msgType == hasCrcMsg {
+		crc := crc32.ChecksumIEEE(buf[5:])
+		expected := binary.BigEndian.Uint32(buf[1:5])
+		if crc != expected {
+			m.logger.Printf("[WARN] memberlist: Got invalid checksum for UDP packet: %d, %d", crc, expected)
```
slackpad: Print as hex is probably easier to interpret.

net.go
```diff
@@ -613,11 +637,11 @@ func (m *Memberlist) sendMsg(to net.Addr, msg []byte) error {
 	compound := makeCompoundMessage(msgs)
 
 	// Send the message
-	return m.rawSendMsgUDP(to, compound.Bytes())
+	return m.rawSendMsgUDP(to, node, compound.Bytes())
 }
 
 // rawSendMsgUDP is used to send a UDP message to another host without modification
```
slackpad: Let's add a comment about it being ok for the node to be `nil`.

net.go
```diff
@@ -631,6 +655,16 @@ func (m *Memberlist) rawSendMsgUDP(to net.Addr, msg []byte) error {
 		}
 	}
 
+	// Add a CRC to the end of the payload if the recipient understands
+	// ProtocolVersion >= 5
+	if node != nil && node.PMax >= 5 {
+		crc := crc32.ChecksumIEEE(msg)
+		sum := make([]byte, 5)
```
slackpad: `sum` is kind of a confusing name, how about `header`?

memberlist_test.go
kyhavlov: This test was already failing in master, so I had it skipped before getting around to fixing it in this PR

slackpad:
LGTM

PR #100, opened 2016-12-13
Change suspicion timeout function to be smooth instead of stepwise

### LINKED ISSUE #98: suspicion timeouts are quantized by whole-number multiples of Log10(cluster size)
The current logic (in util.go:suspicionTimeout) for computing the suspicion timeouts is:

	nodeScale := math.Ceil(math.Log10(float64(n + 1)))
	timeout := time.Duration(suspicionMult) * time.Duration(nodeScale) * interval

The Ceil operator causes nodeScale to always be a whole-number multiple of Log10(n+1), with the result that the timeout increases as a step function:

```
           n < 10    =>  timeout = 1 x suspicionMult x interval
     10 <= n < 100   =>  timeout = 2 x suspicionMult x interval
    100 <= n < 1000  =>  timeout = 3 x suspicionMult x interval
```

This change:

  nodeScale := math.Max(1.0, math.Log10(math.Max(1.0, float64(n))))

would make it rise smoothly (but still not dropping below 1x if n < 10):

```
      n = 5       =>  timeout = 1 x suspicionMult x interval
      n = 10      =>  timeout = 1 x suspicionMult x interval
      n = 50      =>  timeout = 1.7 x suspicionMult x interval
      n = 100     =>  timeout = 2.0 x suspicionMult x interval
      n = 500     =>  timeout = 2.7 x suspicionMult x interval
      n = 1000    =>  timeout = 3.0 x suspicionMult x interval
```

It still guards against n == 0, which I believe was the intention of the +1 in (n + 1).

kyhavlov:
Resolves #98.

### DIFF
--- a/util.go
+++ b/util.go
@@ -155,8 +155,9 @@ func randomOffset(n int) int {
 // suspicionTimeout computes the timeout that should be used when
 // a node is suspected
 func suspicionTimeout(suspicionMult, n int, interval time.Duration) time.Duration {
-	nodeScale := math.Ceil(math.Log10(float64(n + 1)))
-	timeout := time.Duration(suspicionMult) * time.Duration(nodeScale) * interval
+	nodeScale := math.Max(1.0, math.Log10(math.Max(1.0, float64(n))))
+	// multiply by 1000 to keep some precision because time.Duration is an int64 type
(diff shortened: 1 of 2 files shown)


slackpad:
LGTM

PR #103, opened 2017-01-09
preference a privateIP address over loopback addresses

### LINKED ISSUE #102: util.privateBlocks contains loopback address range
When determining the advertiseAddr the IP of each network interface is examined to see if it is a privateIP.

The set of privateIp ranges that the addresses are evaluated against include the loopback address range (127.0.0.0/8).  

When iterating over the interface addresses, the loopback address (127.0.0.1) is always returned first on my system (Ubuntu 14.04- 3.13.0-100-generic #147-Ubuntu).

As a result, 127.0.0.1 is always selected as the privateIP address to advertise to peers, which results in the peers being unable to communicate with each other.

I really think that the loopback range should be removed from the privateBlocks or at the very least moved to the end of the slice so that all other private Ip ranges are checked first.

woodsaj:
- prefer private networks over link-local and loopback addresses.
- fixes #102

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -371,14 +371,19 @@ func (m *Memberlist) setAlive() error {
 				if !IsPrivateIP(ip.String()) {
 					continue
 				}
-
 				advertiseAddr = ip
-				break
+				// if the ip is a loopback addr, keep trying other interface addresses.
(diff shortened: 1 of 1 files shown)


slackpad:
Closing this in favor of #108.

PR #106, opened 2017-01-28
resetNodes() should respect gossip to dead

bosconi:
https://github.com/hashicorp/memberlist/pull/99 introduced `Config.GossipToTheDeadTime`.

`resetNodes()` is called when a node gets to the end of its randomized list of nodes to gossip with. It should not remove dead nodes if they entered the dead state less than `Config.GossipToTheDeadTime` ago.

### DIFF
--- a/state_test.go
+++ b/state_test.go
@@ -862,8 +862,7 @@ func TestMemberList_ResetNodes(t *testing.T) {
 	d := dead{Node: "test2", Incarnation: 1}
 	m.deadNode(&d)
 
-	oldGossipToTheDeadTime := m.config.GossipToTheDeadTime
-	m.config.GossipToTheDeadTime = 3 * time.Second
+	m.config.GossipToTheDeadTime = 200 * time.Millisecond
 	m.resetNodes()
 	if len(m.nodes) != 3 {
 		t.Fatalf("Bad length")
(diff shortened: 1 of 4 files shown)


state_test.go
```diff
@@ -862,13 +862,26 @@ func TestMemberList_ResetNodes(t *testing.T) {
 	d := dead{Node: "test2", Incarnation: 1}
 	m.deadNode(&d)
 
+	oldGossipToTheDeadTime := m.config.GossipToTheDeadTime
```
slackpad: No need to save and restore this - the scope here is just this one test.

state_test.go
```diff
@@ -862,13 +862,26 @@ func TestMemberList_ResetNodes(t *testing.T) {
...
+	m.config.GossipToTheDeadTime = 3 * time.Second
```
slackpad: Let's set this to something like `200 * time.Millisecond` so the test doesn't take too long.

bosconi:
Makes sense. Done.

state_test.go
```diff
@@ -862,6 +862,16 @@ func TestMemberList_ResetNodes(t *testing.T) {
...
+	m.config.GossipToTheDeadTime = 200 * time.Millisecond
+	m.resetNodes()
+	if len(m.nodes) != 3 {
+		t.Fatalf("Bad length")
+	}
+	if _, ok := m.nodeMap["test2"]; !ok {
+		t.Fatalf("test2 should not be unmapped")
+	}
+
+	time.Sleep(5 * time.Second)
```
slackpad: We can knock this one down, too.

PR #107, opened 2017-01-28
gossip() should send a single message without compound message wrapper

bosconi:
`makeCompoundMessage()` should only be called when there are multiple messages to send together. `sendMsg` and `probeNode()` correctly guard for this, but `gossip()` currently does not.

Note that no unit test change accompanies this fix, as there does not appear to be a unit test that can detect this difference, or a way to add one without intrusively modifying the methods in this area. The fix was tested by temporarily inserting additional logging and running on a 100 node instance of Consul, with injected anomalies. The single message code path was observed to be taken around 55% of the time.

### DIFF
--- a/state.go
+++ b/state.go
@@ -496,13 +496,19 @@ func (m *Memberlist) gossip() {
 			return
 		}
 
-		// Create a compound message
-		compound := makeCompoundMessage(msgs)
-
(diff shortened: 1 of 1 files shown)


slackpad:
Thanks!

PR #118, opened 2017-04-14
Add String() method to Node type

mattbostock:
So that we can easily print an identifier for a node, e.g. in logs.

This also means that you can use `list.Members()` in a printf statement
and get a human-readable list of members, e.g.:

    [foo,bar,baz]


slackpad:
Thanks!

PR #120, opened 2017-05-25
Add more tests on ResolveAddr

meyskens:
I have noticed while ResolveAddr accepts adresses without ports on it there was no test on it.


slackpad:
Thanks!

PR #125, opened 2017-06-22
Optimize Broadcast handling

fcrisciani:
- Introduced a flag to trigger queue reordering
- avoid queue copy on broadcast insertion, just swap the element and mark the queue for reordering if needed
- on removal of the message during transmission, keep the queue ordering so that an end sorting is not needed, but also be sure to not skip one message.
In the previous logic: q.bcQueue[i], q.bcQueue[n-1] = q.bcQueue[n-1], nil
so q.bcQueue[i] has the last element of the queue but won't be used because the next cycle i would take i--.

Signed-off-by: Flavio Crisciani <flavio.crisciani@docker.com>

### DIFF
--- a/queue.go
+++ b/queue.go
@@ -21,6 +21,8 @@ type TransmitLimitedQueue struct {
 
 	sync.Mutex
 	bcQueue limitedBroadcasts
+
+	needSort bool // The queue need to be sorted
 }
 
 type limitedBroadcast struct {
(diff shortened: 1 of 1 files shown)


fcrisciani:
@slackpad WDYT?

fcrisciani:
ping @kyhavlov

fcrisciani:
anyone that can give a comment on this?

fcrisciani:
Actually looking better at the code, the last sorting is still needed in case the size does not allow to process all the messages. Also the swap is fine because is an element already used and will be reordered by the next sort

slackpad:
Sorry we didn't get to this sooner!

PR #127, opened 2017-06-30
Use the generic net interfaces instead of the concrete types

avinoamr:
Current NetTransport implementation has hard-coded references to the concrete connection types [net.TCPListener](https://golang.org/pkg/net/#TCPListener) and [net.UDPConn](https://golang.org/pkg/net/#UDPConn), instead of the generic interfaces that they implement: [net.Listener](https://golang.org/pkg/net/#Listener) and [net.PacketConn](https://golang.org/pkg/net/#PacketConn) respectively.

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -43,8 +43,8 @@ type NetTransport struct {
 	streamCh     chan net.Conn
 	logger       *log.Logger
 	wg           sync.WaitGroup
-	tcpListeners []*net.TCPListener
-	udpListeners []*net.UDPConn
+	tcpListeners []net.Listener
+	udpListeners []net.PacketConn
 	shutdown     int32
 }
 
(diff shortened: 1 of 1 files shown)


avinoamr:
I also would like to be able to provide my own Listeners, in order to be able to re-use the Memberlist port with other internal communication protocols that I use. I thought of adding the following to the NetTransportConfig:

```go
type NetTransportConfig {
    // ... same as before

    // user-provided listeners. If left nil, the listeners will be created from this structure
    ListenersTCP []net.Listener
    ListenersUDP []net.PacketConn
}
```

What do you think?

slackpad:
Hi @avinoamr thanks for the PR. The type changes make sense.

> I also would like to be able to provide my own Listeners, in order to be able to re-use the Memberlist port with other internal communication protocols that I use.

That looks pretty reasonable, though it's not super clear what this would do - https://github.com/hashicorp/memberlist/blob/v0.1.0/net_transport.go#L125-L169. Unfortunately, memberlist requires all protocols to be on the same port, it's not yet equipped to deal properly with a fully general list of listeners. We've talked internally about extending to support that, but it would be a fairly involved change to do in a backward-compatible way.

avinoamr:
Yeah, I know. You're right. After submitting the PR I noticed that and ended up working around it by wrapping around the NetTransporter and routing some of the connections to my own code if they begin with a certain prefix bytes - otherwise it will continue to memberlist normally. 

So this PR wouldn't help my use-case, but I still think it might be worthwhile to merge just for elegancy.

PR #129, opened 2017-07-17
Reduce LAN min suspicion multiplier

bosconi:
As per the analysis of the performance of the Lifeguard extensions,
described in [Lifeguard : SWIM-ing with Situational
Awareness](https://arxiv.org/abs/1707.00788)


slackpad:
LGTM - I'll pull this in to Serf and Consul shortly.

PR #139, opened 2017-11-16
Moves the RTT start time capture before I/O.

slackpad:
Before this was doing it right after, but if a machine is super busy we can see negative round trip time estimates, which is bad for the Vivaldi algorithm. The rationale before was it was trying to get as close as possible to the I/O, but that wasn't a good assumption.


slackpad:
Got an out-of-band review.

PR #144, opened 2018-02-08
Reference an available SWIM paper link

nstielau:
The Cornell link 404s.  The IEEE link requires payment for the full text, but does provide an abstract, and seems to reasonably be the canonical place for accessing the paper.


slackpad:
Thanks!

PR #150, opened 2018-05-14
Support closed incoming connections channel

aviaoh:
Allow closed incoming connections channel, without failing


aviaoh:
@armon any feedback will be highly appricated 😄

PR #152, opened 2018-05-27
Makefile typo fix .PNONY -> .PHONY

t3hmrman:


t3hmrman:
Absolutely no problem! It was such a small thing that I wondered if there was even a point submitting a PR for it

PR #155, opened 2018-06-21
Integrate hclog

dotpy3:
With this pull request, `memberlist` uses `hclog.Logger` as logging interface.

Here are a few other changes that this commit integrates:
+ The `Config` and `NetTransportConfig` are changed, and now take directly the `hclog.Logger` object. If no object is passed, `newMemberlist` uses `hclog.Default()`.
+ `fromAddress` and `fromConn` are now methods of a struct that embeds `hclog.Logger`.
    + Given that the way those methods are called has changed, tests have also changed.
+ To log errors, I drew inspiration from other projects using `hclog` and added them with the `"err"` field name.

### DIFF
--- a/broadcast.go
+++ b/broadcast.go
@@ -52,7 +52,7 @@ func (m *Memberlist) encodeAndBroadcast(node string, msgType messageType, msg in
 func (m *Memberlist) encodeBroadcastNotify(node string, msgType messageType, msg interface{}, notify chan struct{}) {
 	buf, err := encode(msgType, msg)
 	if err != nil {
-		m.logger.Printf("[ERR] memberlist: Failed to encode message for broadcast: %s", err)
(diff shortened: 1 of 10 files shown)


tyndyll:
Is there any way to move this forward?

dnephin:
You can use `hclog` with memberlist by calling `Logger.StandardLogger` (https://pkg.go.dev/github.com/hashicorp/go-hclog#Logger), and passing that in as the `Logger`:

```go
logger := hclog.New(nil)
logger.StandardLogger(&hclog.StandardLoggerOptions{InferLevels: true})
```

This PR seems to have many merge conflicts, so it would probably be easier to start from scratch. If someone were to do that, it might be preferable to accept an interface that can be satisfied by `hclog` instead.

PR #156, opened 2018-07-10
Fix the error that Errorf call needs 2 args but has 3 args

hackerwins:


banks:
Hey @hackerwins thanks for this, I just hit the same issues and a bunch of others in recent go versions so made an equivalent change in #160 .

Thanks for the contribution!

hackerwins:
@banks Thanks.

PR #159, opened 2018-08-08
Prioritize alive message over other messages

armon:
This PR modifies the way UDP messages are handled to add message priorities. Previously, all messages would be decoded and added FIFO to the handoff queue. Under high pressure, delayed message processing can cause flapping in the cluster. By prioritizing alive messages we minimize the churn by ensuring timely processing and allowing us to ignore older suspect / dead messages.

We add two linked lists, a low and high priority queue. Each queue is LIFO, and we always check the high priority queue for work first. LIFO ordering is preferred to ensure more recent messages are processed first when there is slow processing to avoid churn.

Depends on #158

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -15,6 +15,7 @@ multiple routes.
 package memberlist
 
 import (
+	"container/list"
 	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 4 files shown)


bosconi:
Good stuff. Nice use of LIFO.

PR #160, opened 2018-08-09
Fix tests

banks:
A bunch of our tests here are either broken by changes to recent go versions or are timing dependent and fail most of the time on my machine.

This PR fixes the hard failures with go 1.10.3 (one wouldn't even compile any more due to vet warnings being promoted). And adds a little lighteeeight retry scheme for some others.

This is far from perfect - I ran the whole suite 100 times locally and got about 15/100 runs still fail, but it's way better than it was.

### DIFF
--- a/state_test.go
+++ b/state_test.go
@@ -1767,7 +1767,7 @@ func TestMemberlist_GossipToDead(t *testing.T) {
 		time.Sleep(3 * time.Millisecond)
 
 		if len(ch) < 2 {
-			failf("expected 3 messages from gossip")
+			failf("expected 2 messages from gossip")
 		}
 	})
 }
--- a/state_test.go
+++ b/state_test.go
@@ -667,11 +667,17 @@ func TestMemberList_ProbeNode_Awareness_MissedNack(t *testing.T) {
 		t.Fatalf("took to long to probe, %9.6f", probeTime.Seconds())
 	}
 
-	// We should have gotten dinged for the missed nack.
+	// We should have gotten dinged for the missed nack. Note that the code under
+	// test is waiting for probeTimeMax and then doing some other work before it
+	// updates the awareness, so we need to wait some extra time. Rather than just
+	// add longer and longer sleeps, we'll retry a few times.
 	time.Sleep(probeTimeMax)
-	if score := m1.GetHealthScore(); score != 1 {
-		t.Fatalf("bad: %d", score)
-	}
+	retry(t, 5, 10*time.Millisecond, func(failf func(string, ...interface{})) {
+		if score := m1.GetHealthScore(); score != 1 {
+			failf("expected health score to decrement on missed nack. want %d, "+
+				"got: %d", 1, score)
+		}
+	})
 }
 
 func TestMemberList_ProbeNode_Awareness_OldProtocol(t *testing.T) {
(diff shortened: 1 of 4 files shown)


state_test.go
```diff
@@ -1729,15 +1760,16 @@ func TestMemberlist_GossipToDead(t *testing.T) {
...
 
-	for i := 0; i < 2; i++ {
-		select {
-		case <-ch:
-		case <-time.After(50 * time.Millisecond):
-			t.Fatalf("timeout")
+	retry(t, 5, 10*time.Millisecond, func(failf func(string, ...interface{})) {
+		m1.gossip()
+
+		time.Sleep(3 * time.Millisecond)
```
mkeeler: Are we expecting 1 message per Millisecond?

If so couldn't we sleep for `3333 * time.Microsecond` or something with a little extra to increase reliability of the tests?

I have very little context here so I could be wrong.

state_test.go
```diff
@@ -1765,16 +1797,16 @@ func TestMemberlist_PushPull(t *testing.T) {
 	a2 := alive{Node: addr2.String(), Addr: ip2, Port: 7946, Incarnation: 1}
 	m1.aliveNode(&a2, nil, false)
 
-	// Gossip should send all this to m2
-	m1.pushPull()
+	// Gossip should send all this to m2. It's UDP though so retry a few times
+	retry(t, 5, 10*time.Millisecond, func(failf func(string, ...interface{})) {
+		m1.pushPull()
+
+		time.Sleep(3 * time.Millisecond)
```
mkeeler: ditto

state_test.go
```diff
@@ -1729,15 +1760,16 @@ func TestMemberlist_GossipToDead(t *testing.T) {
...
-		case <-ch:
-		case <-time.After(50 * time.Millisecond):
-			t.Fatalf("timeout")
+	retry(t, 5, 10*time.Millisecond, func(failf func(string, ...interface{})) {
+		m1.gossip()
+
+		time.Sleep(3 * time.Millisecond)
+
+		if len(ch) < 2 {
+			failf("expected 3 messages from gossip")
```
freddygv: This should be "expected 2 messages from gossip"

state_test.go
```diff
@@ -1680,15 +1686,40 @@ func TestMemberlist_Gossip(t *testing.T) {
 	a3 := alive{Node: "172.0.0.1", Addr: []byte{172, 0, 0, 1}, Incarnation: 1}
 	m1.aliveNode(&a3, nil, false)
 
-	// Gossip should send all this to m2
-	m1.gossip()
+	// Gossip should send all this to m2. Retry a few times because it's UDP and
+	// timing and stuff makes this flaky without.
+	retry(t, 5, 10*time.Millisecond, func(failf func(string, ...interface{})) {
+		m1.gossip()
```
freddygv: Should this be outside of the retry loop? Seems like we should be retrying to wait for the arrival of delayed messages and not to also broadcast more times. (Same comment for L1735)

state_test.go
banks: No we are expecting them all to be delivered, this is just giving go scheduler and network stack a bit of space otherwise it would likely fail every single time on the first retry. Could remove it and let it always fail and be retried but seemed better to leave it with a chance of passing!

Issue #131, opened 2017-08-15
Member list has flaky tests 


preetapan:
Tests timeout intermittently, needs to be more robust.

PR #166, opened 2018-10-24
Unflake a bunch of tests.

rboyer:
* TestMemberlist_EncryptedGossipTransition was rewritten to actually test a recommended upgrade path, instead of testing a crazy split-brain situation that would fail the test roughly half of the time.
* Forced tests to defer a call to Memberlist.Shutdown to avoid bleedover.
* Forced each test to pick a random port for its cluster to avoid bleedover.
* Got tests to pass the race detector.

### DIFF
--- a/memberlist_test.go
+++ b/memberlist_test.go
@@ -34,11 +34,14 @@ func getBindAddr() net.IP {
 }
 
 func testConfig(tb testing.TB) *Config {
+	tb.Helper()
+
 	config := DefaultLANConfig()
 	config.BindAddr = getBindAddr().String()
 	config.Name = config.BindAddr
+	config.BindPort = 0 // choose free port
 	if tb != nil {
-		config.Logger = testLogger(tb)
+		config.Logger = testLoggerWithName(tb, config.Name)
 	}
 	return config
 }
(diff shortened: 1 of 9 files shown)


memberlist_test.go
```diff
@@ -1316,197 +1364,350 @@ func TestMemberlist_conflictDelegate(t *testing.T) {
...
+		if err := isPortFree(t, addr, port); err != nil {
+			failf("%s port is not yet free", m.config.Name)
+		}
+	})
+}
 
-	// Check the hosts
-	if len(m3.Members()) != 2 {
-		t.Fatalf("should have 2 nodes! %v", m3.Members())
+// This test should follow the recommended upgrade guide:
```
rboyer: The prior version of this test technically _worked_ but was testing an undesirable situation that isn't part of the gossip encryption upgrade guide.  There were 4 nodes (same as the new test):
* A: no encryption
* B: encryption, no-verify
* C: encryption, verify-out
* D: encryption, verify-in, verify-out

There would only ever be 2 full members of the pool at each iteration of the test, but the Left members A and B would remain gossiping.  As A was incapable of understanding encrypted messages, both C and D could not directly communicate with that node, and A had to rely upon gossip forwarded from B, which was still up.

The new version of the test instead tries to follow the spirit of the Official Procedure and manipulates a two-node cluster, transitioning it to the various approved states so that the behavior is more regular and doesn't cause any full or partial network partitions that make the test even more timing dependent.

rboyer:
This should mostly fix #131

PR #168, opened 2018-10-29
Add go.mod

mitchellh:
This adds a `go.mod` file. `go test` passes.

Its likely/possible we may have to change the minimum version in the future as downstreams adopt Go modules. But for now this is a working set and it will have no effect on downstreams that do NOT use go modules (all our projects, currently).

This also updates `.travis.yml` to use Go modules for tests.


vtolstov:
please, merge it

mitchellh:
Okay 😄 

I think this is fine, the existence of a go.mod won't break anything on its own, as the original post I wrote said. We can adjust the minimum versions as necessary.

PR #169, opened 2018-12-17
added back-off to accept loop to avoid a tight loop

cgbaker:


net_transport.go
```diff
@@ -221,16 +221,34 @@ func (t *NetTransport) Shutdown() error {
...
+			} else {
+				loopDelay *= 2
+			}
+
+			if loopDelay > maxDelay {
+				loopDelay = maxDelay
+			}
+
 			t.logger.Printf("[ERR] memberlist: Error accepting TCP connection: %v", err)
+			time.Sleep(loopDelay)
```
banks: The behaviour here is that we will delay `Shutdown` for up to `maxDelay` in case an accept error has just occured.

Given `maxDelay` is one second that seems fine and preferable to changing the shutdown mechanism to close a chan for example.

Is it worth adding a comment to the `maxDelay` const above just to note that assumption? e.g. in case anyone later thought it prudent to increase that value without noticing it would hard the shutdown responsiveness eve further?

net_transport.go
```diff
@@ -221,16 +221,34 @@ func (t *NetTransport) Shutdown() error {
...
 			continue
 		}
+		// No error, reset loop delay
+		loopDelay = 0
```
preetapan: Is this new behavior unit testable in transport_test.go?

net_transport.go
cgbaker: will add a comment. thanks!

net_transport.go
cgbaker: okay, test added... let me know what you think. i struggled a bit to find a good entrypoint for observing the side-effects on this one.

transport_test.go
```diff
@@ -133,3 +136,51 @@ func TestTransport_Send(t *testing.T) {
...
+	numCalls *int32
+}
+
+func (tw testCountingWriter) Write(p []byte) (n int, err error) {
+	atomic.AddInt32(tw.numCalls, 1)
+	tw.t.Log("countingWriter", string(p))
+	return len(p), nil
+}
+
+func TestTransport_TcpListenBackoff(t *testing.T) {
```
preetapan: test looks good but I would add more comments around the assertion or a block comment about the methodology (along the lines that you are verifying that there are an expected number of retries given the current max delay / base delay values)

transport_test.go
```diff
@@ -133,3 +136,51 @@ func TestTransport_Send(t *testing.T) {
...
+
+	c := make(chan struct{})
+	go func() {
+		defer close(c)
+		transport.wg.Wait()
+	}()
+	select {
+	case <- c:
+	case <-time.After(1 * time.Second):
+		t.Error("timed out waiting for transport waitgroup to be done after setting shutdown == 1")
```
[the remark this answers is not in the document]
cgbaker: Good point... I sorta pulled that number out of the air, but it is loosely defined by the maximum delay (which happens to be `1s`), so I'll add a little flex and a comment.

transport_test.go
cgbaker: Will do. Thanks!

PR #178, opened 2019-01-24
Avoid to take into account wrong versions of protocols in Vsn.

pierresouchay:
On Consul, sometimes, nodes do send a pMin = pMan = 0 in Vsn
This causes a corruption of the acceptable versions of protocol
and thus requiring version = [0, 1].

After this corruption occurs, all new nodes cannot join anymore, it
then force the restart of all Consul servers to resume normal
operations.

While not fixing the root cause, this patch discards alive nodes
claiming version 0,0,0 and will avoid this breakage.

See https://github.com/hashicorp/consul/issues/3217

### DIFF
--- a/state.go
+++ b/state.go
@@ -896,6 +896,14 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 			},
 			State: stateDead,
 		}
+		if len(a.Vsn) > 5 {
+			state.PMin = a.Vsn[0]
+			state.PMax = a.Vsn[1]
+			state.PCur = a.Vsn[2]
+			state.DMin = a.Vsn[3]
+			state.DMax = a.Vsn[4]
+			state.DCur = a.Vsn[5]
+		}
 
 		// Add to map
 		m.nodeMap[a.Node] = state
--- a/state.go
+++ b/state.go
@@ -850,6 +850,16 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 		return
 	}
 
+	if len(a.Vsn) > 3 {
+		pMin := a.Vsn[0]
+		pMax := a.Vsn[1]
+		pCur := a.Vsn[2]
+		if pMin == 0 || pMax == 0 || pMin > pMax {
+			m.logger.Printf("[WARN] memberlist: Ignoring an alive message for '%s' (%v:%d) because protocol version(s) are wrong: %d <= %d <= %d should be >0", a.Node, net.IP(a.Addr), a.Port, pMin, pCur, pMax)
+			return
+		}
+	}
+
 	// Invoke the Alive delegate if any. This can be used to filter out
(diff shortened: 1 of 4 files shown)


pierresouchay:
cc @Aestek

Aestek:
LGTM. This provides good protection against bad messages as seen in hashicorp/consul#3217

state.go
```diff
@@ -886,6 +896,14 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 			},
 			State: stateDead,
 		}
+		if len(a.Vsn) > 5 {
```
mkeeler: Is this the place where the versions were getting zeroed due to not setting them?

state.go
```diff
@@ -980,6 +995,7 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 			state.DMax = a.Vsn[4]
 			state.DCur = a.Vsn[5]
 		}
+		m.encodeBroadcastNotify(a.Node, aliveMsg, a, notify)
```
mkeeler: Why did this need to get moved. Does something in `encodeBroadcastNotify` use the state's version numbers?

state.go
```diff
@@ -850,6 +850,16 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 		return
 	}
 
+	if len(a.Vsn) > 3 {
```
i0rek: Is it necessary to check if there are more than 3 items in `a.Vsn`?

```suggestion
	if len(a.Vsn) >= 3 {
```

pierresouchay:
A few comments:
For some reason (legacy I suppose), it supports receiving empty Vsn (this is tested in a unit test), but there is no security before those lines:
```

**Comment by pierresouchay** [comment]
@i0rek @mkeeler This bug has been reproduced quite a large time in our infrastructure because we have lots of agents being started/shutdown very often (due to poor lifecycle of somee of our machines).

I have been able to reproduce it quite easily with very recent Consul version (here 3e299c0192b5277b938d99ff0fadfba45f2835a9 )

1. Start a non-patched server :
```
consul agent -data-dir $PWD/server1 -node server1 -server -datacenter testDC -bootstrap-expect 1 -serf-lan-port 9500
```

2. Apply a very small patch to file `vendor/github.com/hashicorp/memberlist/memberlist.go`, by adding `time.Sleep(1 * time.Second)` at line 202, just before line `if err := m.setAlive(); err != nil {`

3. Run the following command:
```
while true; do consul agent -data-dir $PWD/client1 -node client1 -join 192.168.3.31:9500  -serf-lan-port 10300 -client 127.0.0.2 -datacenter=testDC -hcl leave_on_terminate=false & sleep 1.2 ; kill -9 %1; sleep .1; done
```

In a few seconds, on the server:
```

banks:
LGTM! Thanks!

Issue #134, opened 2017-09-26
Data Races


magiconair:
@slackpad a data race on `Incarnation`. Maybe you have an idea on how to address this:

```
WARNING: DATA RACE
Write at 0x00c424b902d0 by goroutine 5112:
  github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist.(*Memberlist).refute()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist/state.go:807 +0x7f
  github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist.(*Memberlist).suspectNode()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist/state.go:1033 +0xa41
  github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist.(*Memberlist).handleSuspect()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist/net.go:518 +0x297
  github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist.(*Memberlist).packetHandler()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist/net.go:384 +0x360

Previous read at 0x00c424b902d0 by goroutine 168:
  github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist.(*Memberlist).Leave()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/vendor/github.com/hashicorp/memberlist/memberlist.go:580 +0x269
  github.com/hashicorp/consul/vendor/github.com/hashicorp/serf/serf.(*Serf).Leave()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/vendor/github.com/hashicorp/serf/serf/serf.go:690 +0x2fc
  github.com/hashicorp/consul/agent/consul.(*Client).Leave()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/agent/consul/client.go:180 +0xf9
  github.com/hashicorp/consul/agent/consul.TestLeader_LeftMember()
      /Users/frank/go/consul/src/github.com/hashicorp/consul/agent/consul/leader_test.go:178 +0x25b
  testing.tRunner()
      /Users/frank/go1.9/src/testing/testing.go:746 +0x16c

Goroutine 5112 (running) created at:
[...]

PR #179, opened 2019-01-24
eliminate one of the data races on access to node data

### LINKED ISSUE #113: Race found with Node.PMax
Hi hashis,

Found this data race while using `memberlist` as of commit 23ad4b7d7b38496cd64c241dfd4c60b7794c254a :
```==================
WARNING: DATA RACE
Read at 0x00c4211acf49 by goroutine 660:
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).rawSendMsgUDP()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/net.go:662 +0x63b
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).gossip()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:509 +0x733
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).(github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.gossip)-fm()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:94 +0x41
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).triggerFunc()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:118 +0x159

Previous write at 0x00c4211acf49 by goroutine 705:
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).aliveNode()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:952 +0xef2
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).handleAlive()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/net.go:558 +0x32f
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).udpHandler()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/net.go:410 +0x308

Goroutine 660 (running) created at:
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.(*Memberlist).schedule()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/state.go:94 +0x423
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.Create()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/memberlist.go:164 +0xcd
  github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf.Create()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf/serf.go:398 +0x13d5
  github.com/fastly/checkup/cluster.New()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/cluster/cluster.go:126 +0x6e0
  github.com/fastly/checkup/checkup.New()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/checkup/checkup.go:206 +0x13a7
  github.com/fastly/checkup/test.StartCheckup()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/test/helper.go:62 +0x36f
  command-line-arguments_test.startManyCheckupsAndFakeVarnishes()
      /home/vagrant/devel/go-workspace/checkup/fakevarnish/scale_test.go:333 +0x534
  command-line-arguments_test.TestManyCheckupsWithDifferentBackends()
      /home/vagrant/devel/go-workspace/checkup/fakevarnish/scale_test.go:166 +0x20f
  testing.tRunner()
      /home/vagrant/devel/go-workspace/go/src/testing/testing.go:657 +0x107

Goroutine 705 (running) created at:
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.newMemberlist()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/memberlist.go:146 +0xde0
  github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist.Create()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/memberlist/memberlist.go:156 +0x3c
  github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf.Create()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/vendor/github.com/hashicorp/serf/serf/serf.go:398 +0x13d5
  github.com/fastly/checkup/cluster.New()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/cluster/cluster.go:126 +0x6e0
  github.com/fastly/checkup/checkup.New()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/checkup/checkup.go:206 +0x13a7
  github.com/fastly/checkup/test.StartCheckup()
      /home/vagrant/devel/go-workspace/src/github.com/fastly/checkup/test/helper.go:62 +0x36f
  command-line-arguments_test.startManyCheckupsAndFakeVarnishes()
      /home/vagrant/devel/go-workspace/checkup/fakevarnish/scale_test.go:333 +0x534
  command-line-arguments_test.TestManyCheckupsWithDifferentBackends()
      /home/vagrant/devel/go-workspace/checkup/fakevarnish/scale_test.go:166 +0x20f
  testing.tRunner()
      /home/vagrant/devel/go-workspace/go/src/testing/testing.go:657 +0x107
==================
```

I'm not sure if `nodeLock` (https://github.com/hashicorp/memberlist/blob/902b55b5d29084b62aa0d7989ed7f89ee7e9230e/memberlist.go#L47) is intended to protect just `nodeMap` in `aliveNode()` or also the `nodeState ` / `Node`s from the map and their data (such as `PMax`).

Looks like potential races also in `probeNode`:
https://github.com/hashicorp/memberlist/blob/master/state.go#L326
https://github.com/hashicorp/memberlist/blob/master/state.go#L347

seebs:
This particular data race hits us a lot in CI testing. The
underlying issue is that access to the contents of m.nodeMap
is controlled by a mutex, but rawSendMsgPacket is using the
mutex only to grab the pointer -- then accessing the contents
of the pointer without the mutex held. This means a data race
if something else grabs the mutex and modifies the contents
of the pointer as well.

In this particular case, access is read-only, so we copy the
pointed-to thing into a local. Less efficient, but also less
of a data race.

(Probably)
Fixes #113, #134 
(Well, not all of #134, but this was at least one of them.)

### DIFF
--- a/net.go
+++ b/net.go
@@ -678,10 +678,14 @@ func (m *Memberlist) rawSendMsgPacket(addr string, node *Node, msg []byte) error
 		}
 		m.nodeLock.RLock()
 		nodeState, ok := m.nodeMap[toAddr]
-		m.nodeLock.RUnlock()
 		if ok {
-			node = &nodeState.Node
+			// if we don't copy this data, we'll be accessing the
+			// pointed-to thing without the mutex held, which means
+			// data races, usually against aliveNode() in state.go.
+			nodeCopy := nodeState.Node
+			node = &nodeCopy
 		}
+		m.nodeLock.RUnlock()
 	}
 
 	// Add a CRC to the end of the payload if the recipient understands
--- a/util.go
+++ b/util.go
@@ -149,6 +149,16 @@ OUTER:
 		// Append the node
 		kNodes = append(kNodes, node)
 	}
+	kNodeStorage := make([]nodeState, len(kNodes))
+	// Copy the contents of the node pointers out to local storage, and
+	// return pointers to that, so we avoid races.
+	for i, node := range kNodes {
+		kNodeStorage[i] = *node
+		metaCopy := make([]byte, len(node.Meta))
+		copy(metaCopy, node.Meta)
+		kNodeStorage[i].Meta = metaCopy
+		kNodes[i] = &kNodeStorage[i]
+	}
 	return kNodes
 }
 


net.go
```diff
@@ -678,10 +678,15 @@ func (m *Memberlist) rawSendMsgPacket(addr string, node *Node, msg []byte) error
...
 		if ok {
-			node = &nodeState.Node
+			// if we don't copy this data, we'll be accessing the
+			// pointed-to thing without the mutex held, which means
+			// data races, usually against aliveNode() in state.go.
+			nodeCopy := nodeState.Node
+			node = &nodeCopy
 		}
+		m.nodeLock.RUnlock()
+
```
JensRantil:
```suggestion
```

seebs: yeah, probably. the extra blank line wasn't intentional.

jaffee:
Ping on this - the races are real and hit us occasionally. Thanks!

seebs:
pushed update to drop the spare newline. my branch should now be on 7e5a340e.

seebs:
So I got hit by one of them again, and studied the code, and realized that the branch I fixed wasn't probably the relevant branch -- it was in the case where `rawSendMsgPacket` *is* given a node pointer. On further study, it turns out there's a general pattern of using `kRandomNodes` to collect some nodes, then operate on them -- but while `kRandomNodes` is usually called with the node mutex, the resulting operations aren't.

Modifying `kRandomNodes` to copy the nodes it is returning seems to help. If I run the test that was hitting this 10 times without this patch, I "probably" get a data race. If I run it 100 times with, I don't see any races.

```
Write at 0x00c0005b7249 by goroutine 441:
  github.com/hashicorp/memberlist.(*Memberlist).aliveNode()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/state.go:999 +0xf16
  github.com/hashicorp/memberlist.(*Memberlist).mergeState()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/state.go:1187 +0x3b2
  github.com/hashicorp/memberlist.(*Memberlist).mergeRemoteState()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/net.go:1071 +0x642
  github.com/hashicorp/memberlist.(*Memberlist).handleConn()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/net.go:264 +0xa36

Previous read at 0x00c0005b7249 by goroutine 439:
  github.com/hashicorp/memberlist.(*Memberlist).rawSendMsgPacket()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/net.go:693 +0x8e3
  github.com/hashicorp/memberlist.(*Memberlist).gossip()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/state.go:534 +0x650
  github.com/hashicorp/memberlist.(*Memberlist).gossip-fm()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/state.go:489 +0x41
  github.com/hashicorp/memberlist.(*Memberlist).triggerFunc()
      /go/pkg/mod/github.com/pilosa/memberlist@v0.1.4-0.20190408132233-ff8741fd3108/state.go:135 +0x179
```

This is a sample of the trace, and the underlying issue is that `gossip` is grabbing a list of nodes and calling `rawSendMsgPacket`, which checks the state of the node it's invoked on, while `aliveNode` may be modifying that node's state. With the patch, `gossip` grabs the old node states, and operates on them, and if updates come in it misses them, but it's no longer a data race. (The other alternative might be to, say, request a lock again before evaluating the contents of the node, but I assume that if it were practical to run the entire operation with a read lock held, it would already be being done?)

seebs:
Oh, interesting. That test failure makes me think that the code *does* want to get those updates... In which case I'm not at all sure what a correct fix would look like.

banks:
Hey thanks for the PR @seebs there are several races caught by the detector here which is certainly not ideal. One issue with fixing them though is validating the performance hit caused e.g. by copying lots of stuff.

For context we know of users who run Consul clusters with over 10k nodes in one Gossip pool. For them any operation like PushPull that has to operate in O(n) over the number of nodes is already problematic so the performance impact of copying all the node states every time we pick a new node to probe (at least every second) would likely be prohibitive in that environment.

Do we know if anyone has observed actual incorrect behaviour due to these races or is it "just" being picked up by the race detector. I say "just" because I still think it's important to not have data races at all even if they don't seem to cause problems but it's important in deciding how to trade off against performance and time spent finding a solution that appeases the race detector (and is correct in operation) without impacting performance.

All that said, I don't _know_ this would be an issue to merge as-is but I suspect it would so without some in-depth performance analysis (by us or anyone else who is inclined to do it) it's risky for us! Testing these things at scale is pretty time consuming as you can imagine. A microbenchmark of the code paths changed might at least give some basic data we can use to get an idea of the potential issue.

If you have any more insights on that or are interested in finding possible solutions that might impact performance less we're interested to hear. We hope to get to investigating these races more at some point however as always it's a prioritisation problem - lots of work to do vs. not clear if it's actually impacting users in production (outside of CI complaining which is also a legit problem but a different one!).

seebs:
Yeah, the performance hit is definitely a problem with this fix. I'm not sure what alternatives exist. We have a similar thing in our code base where I implemented it as copy-on-write; when things get returned by something like kRandomNodes, they get a "frozen" flag set; anything that wants to change them, thereafter, will actually replace them with a new object. So the object gets copied if and only if it's being written after being copied out by something that expects it to be read outside the lock. This design is working quite nicely for us, so far.

I don't know about actual incorrect behavior due to the races, but anecdotally, I think our incidence rate of Weird Gossip Failures We Can't Directly Explain went down after we changed this, even though the motivation for changing it was just a ~1-2% failure rate of any and all CI tests using `-race`.

banks:
Interesting, thanks for sharing your experience. Just to be sure - you
deployed your patch here in production? What kind of size is the cluster if
you can share that? Any noticeable increase in CPU from your processes?

seebs:
I haven't looked at "real" production systems, and I've only been doing small clusters. It's probably not-zero, but it's not large enough to be obvious... But it's pretty obvious to me that it's going to be a potential problem. For *my* purposes, "fix the problem so I stop getting CI failures" is a good choice. For larger clusters, I'd probably want to actually implement something like copy-on-write.

The basic idea is: things like kRandomNodes, which will return nodeStates to things outside a lock, set a "frozen" flag. Anything which would be writing to the data that gets read outside the lock checks for that flag, and if it's set, then copies that specific node, modifies the new one, and stores the new pointer in whatever list it's looking at. This works in our system, because at any given point, usually only one thing has a pointer to a given object of the type in question at once, unless it's a transient "copy" that's being used to continue accessing the data outside a lock.

For read-heavy workloads, things get marked frozen and stay that way, for write-heavy workloads, they don't get frozen, we only copy them a lot when there's both reads and writes. But with a large cluster, the assumption would be that, while nodes change state fairly often, they don't *all* change state all that often, so while there'd still be a lot of new nodeState objects coming into existence, it wouldn't be "all k of them, unconditionally, every time we want kRandomNodes".

The other concern I've got is related to the CI problem: Even if the data races aren't actually resulting in corruption (which I'd be slightly skeptical of; it seems like it's possible to get inconsistent reads in a way that would make nonsensical choices happen), they're resulting in not running with race detector enabled, meaning *other* data races are *also* going undetected. Keeping stuff race-free lets us use potentially important tools.

banks:
Thanks.

I totally agree getting it race free is important even just for CI but also
because it's dangerous to leave it racey!

seebs:
So, long story short, I was talking about a race condition thing elsewhere, and I got to thinking.

`sync/atomic` is safe against data races, sort of. It's not safe in that it doesn't provide guarantees that you don't get incoherent data if you read multiple values, but it's safe in that it's not actual full on data races that the race detector cares about.

But it turns out, there's not that many accesses to these nodes, and in practice, I don't think I ever saw one of these races hit where the race didn't involve the read in rawSendMsgPacket. What's that reading from the node? So far as I can tell, the only thing it ever reads is `node.PMax`.

In general, the pointers themselves don't really become invalidated; they're added to m.nodeMap only if no previous state pointer existed, otherwise the state gets updated in-place. And everything is then modifications throught that state pointer.

So... It'd probably be safe to do these with atomic ops. Except now that I look, the fields in question are `uint8` and there's no 8-bit atomic ops in sync/atomic, just 32/64/pointer.

But I've also realized that it's quite possible that those values never actually change in real-world situations. After all, it's quite possible that a given node continues supporting the same protocol versions for its entire functional lifespan, or at least that updates to memberlist are rare. And parallel reads aren't data races. So if those all became

```
if state.PMax != a.Vsn[1] {
    state.PMax = a.Vsn[1]
}
```

it'd probably never trigger again. Or at least, hardly ever.

banks:
It seems like the simplest solution that avoids copies would be to change all call sites that update nodes from this map to do so via a central method that takes the lock. I've not looked through the code to see how many places that is but from your discussion @seebs it seems to be only a few right? Is there some reason that is going to be a difficult fix? I don't think lock contention is too much of a problem given that all the reads are already contending on the same lock.

seebs:
I think everything updating the nodes already has a lock. The race condition is against reads of lists returned from kRandomNodes, I think? And those lists are around for a while after they're obtained; the problem is that the list is obtained under the lock, but then the pointers are dereferenced outside it, which is why my first patch was just to make copies. (Slow but almost certainly correct.)

banks:
Ah got it. I'm not dead against copying provided we are confident the
performance cost isn't significant when you have large clusters.

The rate of updates remains reasonably constant (push pull increases with a
larger cluster but logarithmically rather than linearly but regular gossip
rate doesn't). The killer is the linear growth in number of nodes in the
list. If we could simulate/test with ~10k nodes and show CPU about the same
with copying I'd be fine merging it.

Copy on write would also be a valid solution which might not be too hard to
build depending on how many different code paths perform writes. If they
are all taking lock already then they could just copy and update and
replace in the map under lock and all would be happy presumably - the
readers holding pointers to the old struct would still be using a stale
value but that's OK and already true, but they wouldn't be dereferencing a
pointer to a struct that might later be modified...

Apologies this is all pretty hand wavy as I have not had time to dive into
the code to remind myself of the details!

PR #189, opened 2019-03-14
Allow a dead node's name to be taken by a new node

kyhavlov:
This change goes along with hashicorp/consul#5485 to allow a dead node's name to be reclaimed by a new node. There's more context in that PR but the summary is that a dead node could get stuck holding onto it's name when it was recreated with a new address.

### DIFF
--- a/state.go
+++ b/state.go
@@ -930,8 +930,14 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 	} else {
 		// Check if this address is different than the existing node unless the old node is dead.
 		if !bytes.Equal([]byte(state.Addr), a.Addr) || state.Port != a.Port {
+			// If DeadNodeReclaimTime is configured, check if enough time has elapsed since the node died.
+			var canReclaim bool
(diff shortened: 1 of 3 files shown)


scalp42:
@kyhavlov any chance you can make the change @banks is asking? Thanks a lot for the contribution

kyhavlov:
@banks This is ready for a last look with the change you described above.

state.go
```diff
@@ -930,8 +930,14 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 	} else {
 		// Check if this address is different than the existing node unless the old node is dead.
 		if !bytes.Equal([]byte(state.Addr), a.Addr) || state.Port != a.Port {
+			// If DeadNodeReclaimTime is configured, check if enough time has elapsed since the node died.
+			var canReclaim bool
+			if m.config.DeadNodeReclaimTime > 0 && state.StateChange.Add(m.config.DeadNodeReclaimTime).Before(time.Now()) {
```
banks: nitpick: I think this is correct but I'd find it slightly more intuitive to write this as:

```go
canReclaim := (m.config.DeadNodeReclaimTime > 0 && 
  time.Since(state.StateChange) > m.config.DeadNodeReclaimTime)
```

scalp42:
Thanks a lot @kyhavlov

PR #192, opened 2019-04-09
Fix synchronization error for read/write keyring.keys

zuweiliu:
keyring.keys is not always protected by keyring lock, caused race condition. This fix is to always read out the keys using lock, and process on the local copy to avoid race condition

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -81,13 +81,14 @@ func (k *Keyring) AddKey(key []byte) error {
 	}
 
 	// No-op if key is already installed
-	for _, installedKey := range k.keys {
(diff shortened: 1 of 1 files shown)


dnephin:
Thank you for the PR! Is this still an issue? Are there any tests which fail when run with `go test  -race` ? I did not see this in #232 which enables the race detector.

tgross:
Doing some issue/PR cleanup. Because this wasn't linked to #341 I didn't notice that it was still open, and this issue was resolved in https://github.com/hashicorp/memberlist/pull/342.

PR #196, opened 2019-06-24
Added way to block members using CIDR

pierresouchay:
Added 2 new fields in configuration of memberlists to allow/disallow
others machines joining memberlist.

This will allow protecting some machines from joining a cluster and will increase security as well.

Will be useful to implement https://github.com/hashicorp/consul/pull/5916
as suggested my @mkeller

### DIFF
--- a/config.go
+++ b/config.go
@@ -6,10 +6,7 @@ import (
 	"log"
 	"net"
 	"os"
-	"strings"
 	"time"
-
-	multierror "github.com/hashicorp/go-multierror"
 )
 
 type Config struct {
@@ -255,15 +252,10 @@ func ParseCIDRs(v []string) ([]net.IPNet, error) {
 			nets = append(nets, *net)
 		}
 	}
-<<<<<<< HEAD
-	return nets, err
->>>>>>> Added way to block members using CIDR
-=======
 	if !hasErrors {
 		errs = nil
 	}
 	return nets, errs
->>>>>>> Applied suggestions from @rboyer
 }
 
 // DefaultLANConfig returns a sane set of configurations for Memberlist.
@@ -274,7 +266,6 @@ func ParseCIDRs(v []string) ([]net.IPNet, error) {
 // these values are a good starting point when getting started with memberlist.
 func DefaultLANConfig() *Config {
 	hostname, _ := os.Hostname()
-	allowedCidrs, _ := ParseCIDRs([]string{"0.0.0.0/0", "::0/0"})
 	return &Config{
 		Name:                    hostname,
 		BindAddr:                "0.0.0.0",
@@ -308,8 +299,7 @@ func DefaultLANConfig() *Config {
 
 		HandoffQueueDepth: 1024,
 		UDPBufferSize:     1400,
-		CidrsAllowed:      allowedCidrs,
-		CidrsDenied:       nil,
+		CidrsAllowed:      nil, // same as allow all
 	}
 }
 
@@ -329,22 +319,22 @@ func DefaultWANConfig() *Config {
 	return conf
 }
 
-// IpAllowed return an error if access to memberlist is denied
-func (c *Config) IpAllowed(ip net.IP) error {
-	err := fmt.Errorf("%s is not allowed", ip)
+// IPMustBeChecked return true if IPAllowed must be called
+func (c *Config) IPMustBeChecked() bool {
+	return c.CidrsAllowed != nil
+}
+
+// IPAllowed return an error if access to memberlist is denied
+func (c *Config) IPAllowed(ip net.IP) error {
+	if !c.IPMustBeChecked() {
+		return nil
+	}
 	for _, n := range c.CidrsAllowed {
 		if n.Contains(ip) {
-			err = nil
-			break
+			return nil
 		}
 	}
-	for _, n := range c.CidrsDenied {
-		if n.Contains(ip) {
-			err = fmt.Errorf("%s is denied since part of %s", ip, n)
-			break
-		}
-	}
-	return err
+	return fmt.Errorf("%s is not allowed", ip)
 }
 
 // DefaultLocalConfig works like DefaultConfig, however it returns a configuration
--- a/config.go
+++ b/config.go
@@ -1,9 +1,12 @@
 package memberlist
 
 import (
+	"fmt"
 	"io"
 	"log"
+	"net"
 	"os"
+	"strings"
 	"time"
 )
 
@@ -225,9 +228,33 @@ type Config struct {
 	// meaning nodes cannot be reclaimed this way.
 	DeadNodeReclaimTime time.Duration
 
+<<<<<<< HEAD
 	// RequireNodeNames controls if the name of a node is required when sending
 	// a message to that node.
 	RequireNodeNames bool
+=======
+	CidrsAllowed []net.IPNet
+	CidrsDenied  []net.IPNet
+}
+
+// ParseCIDRs return a possible empty list of all Network that have been parsed
+// In case of error, it returns succesfully parsed CIDRs and the last error found
+func ParseCIDRs(v []string) ([]net.IPNet, error) {
+	nets := make([]net.IPNet, 0)
+	var err error
+	if v == nil {
+		return nets, err
+	}
+	for _, p := range v {
+		_, net, err := net.ParseCIDR(strings.TrimSpace(p))
+		if err != nil {
+			err = fmt.Errorf("invalid cidr: %s", p)
+		} else {
+			nets = append(nets, *net)
+		}
+	}
+	return nets, err
+>>>>>>> Added way to block members using CIDR
 }
 
 // DefaultLANConfig returns a sane set of configurations for Memberlist.
@@ -238,6 +265,7 @@ type Config struct {
 // these values are a good starting point when getting started with memberlist.
 func DefaultLANConfig() *Config {
 	hostname, _ := os.Hostname()
+	allowedCidrs, _ := ParseCIDRs([]string{"0.0.0.0/0", "::0/0"})
 	return &Config{
 		Name:                    hostname,
 		BindAddr:                "0.0.0.0",
@@ -271,6 +299,8 @@ func DefaultLANConfig() *Config {
 
 		HandoffQueueDepth: 1024,
 		UDPBufferSize:     1400,
+		CidrsAllowed:      allowedCidrs,
+		CidrsDenied:       []net.IPNet{},
 	}
 }
 
@@ -288,6 +318,24 @@ func DefaultWANConfig() *Config {
 	conf.GossipInterval = 500 * time.Millisecond
 	conf.GossipToTheDeadTime = 60 * time.Second
 	return conf
+}
+
+// IpAllowed return an error if access to memberlist is denied
+func (c *Config) IpAllowed(ip net.IP) error {
+	err := fmt.Errorf("%s is not allowed", ip)
+	for _, n := range c.CidrsAllowed {
+		if n.Contains(ip) {
+			err = nil
+			break
+		}
+	}
+	for _, n := range c.CidrsDenied {
+		if n.Contains(ip) {
+			err = fmt.Errorf("%s is denied since part of %s", ip, n)
+			break
+		}
+	}
+	return err
 }
 
 // DefaultLocalConfig works like DefaultConfig, however it returns a configuration
--- a/config.go
+++ b/config.go
@@ -305,7 +305,7 @@ func DefaultLANConfig() *Config {
 
 		HandoffQueueDepth: 1024,
 		UDPBufferSize:     1400,
-		CidrsAllowed:      nil, // same as allow all
+		CIDRsAllowed:      nil, // same as allow all
 	}
 }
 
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -220,6 +223,28 @@ type Config struct {
...
+func ParseCIDRs(v []string) ([]net.IPNet, error) {
+	nets := make([]net.IPNet, 0)
+	var err error
+	if v == nil {
+		return nets, err
+	}
+	for _, p := range v {
+		_, net, err := net.ParseCIDR(strings.TrimSpace(p))
+		if err != nil {
+			err = fmt.Errorf("invalid cidr: %s", p)
```
Aestek: If this func is intended to parse user config wouldn't it be better to fail if any one of the cidrs is invalid instead of silently ignoring it (even if a warning is printed on startup it can be easily overlooked) ?

net.go
```diff
@@ -205,11 +205,35 @@ func (m *Memberlist) streamListen() {
 	}
 }
 
+// EnsureCanConnect return the IP from a RemoteAddress
+// return error if this client must not connect
+func (m *Memberlist) EnsureCanConnect(from net.Addr) error {
```
Freeaqingme: May I suggest to rename this to EnsureIsAllowed() instead? 'can connect' can also imply that it checks for whether it would succeed. 'isAllowed' would be less ambiguous I think.

config.go
pierresouchay: might be, the user caller can chose it anyway

net.go
pierresouchay: DONE

pierresouchay:
@mkeeler Is it what you had in mind in https://github.com/hashicorp/consul/pull/5916#issuecomment-504181683 ?

pierresouchay:
@mkeeler Could you have a look and tell me if it is Ok to achieve hashicorp/consul#5916 ?

pierresouchay:
@mkeeler DONE as you suggested

config.go
```diff
@@ -263,6 +289,8 @@ func DefaultLANConfig() *Config {
 
 		HandoffQueueDepth: 1024,
 		UDPBufferSize:     1400,
+		CidrsAllowed:      allowedCidrs,
```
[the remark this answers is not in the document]
pierresouchay: Personally I hate the if nil stuff... I am not convinced with gain, but you are the boss, if you want me to change it, I'll do it

config.go
```diff
@@ -263,6 +289,8 @@ func DefaultLANConfig() *Config {
...
+		CidrsDenied:       []net.IPNet{},
```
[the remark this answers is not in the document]
pierresouchay: OK, I'll do it in next patch

config.go
pierresouchay: @rboyer DONE

state.go
```diff
@@ -893,6 +893,11 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 	// store this node in our node map.
 	var updatesNode bool
 	if !ok {
+		errCon := m.config.IpAllowed(a.Addr)
```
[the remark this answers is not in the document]
pierresouchay: @mkeeler DONE in https://github.com/hashicorp/memberlist/pull/196/commits/fc35abeb47505ac33a08685603d44ff35205dc1e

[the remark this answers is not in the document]
pierresouchay: @mkeeler @rboyer Are you Ok with my new patch or do you want me to switch back to `handleConn` ?

pierresouchay:
> Please take a look at my comments and see if I've misunderstood anything about your intent.

I fixed issues you mentioned. Are you Ok with the implementation I made to protect `NotifyAlive`or do you have another suggestion?

net.go
```diff
@@ -568,6 +568,25 @@ func (m *Memberlist) handleSuspect(buf []byte, from net.Addr) {
 	m.suspectNode(&sus)
 }
 
+// EnsureCanConnect return the IP from a RemoteAddress
+// return error if this client must not connect
+func (m *Memberlist) EnsureCanConnect(from net.Addr) error {
```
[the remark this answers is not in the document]
pierresouchay: Yes, will be fixed in the next patch

config.go
```diff
@@ -220,6 +225,34 @@ type Config struct {
...
 	// meaning nodes cannot be reclaimed this way.
 	DeadNodeReclaimTime time.Duration
+
+	CidrsAllowed []net.IPNet
+	CidrsDenied  []net.IPNet
+}
+
+// ParseCIDRs return a possible empty list of all Network that have been parsed
+// In case of error, it returns succesfully parsed CIDRs and the last error found
+func ParseCIDRs(v []string) ([]net.IPNet, error) {
```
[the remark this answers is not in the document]
pierresouchay: I wanted to allow memberlist users (such as Consul) to use it for Parsing their configuration, but I can hide it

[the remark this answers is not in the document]
pierresouchay: Ok, I moved it to tests and do not export it anymore

pierresouchay:
@rboyer Thank you for the review, I think I addressed all your points!

config.go
```diff
@@ -220,6 +222,11 @@ type Config struct {
 	// reclaimed by one with a different address or port. By default, this is 0,
 	// meaning nodes cannot be reclaimed this way.
 	DeadNodeReclaimTime time.Duration
+
+	// If nil, allow any connection (default), otherwise specify all networks
+	// allowed to connect (you must specify IPv6/IPv4 separatly)
```
rboyer:
```suggestion
	// allowed to connect (you must specify IPv6/IPv4 separately)
```

config_test.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, ip := range tests {
+		localV4 := net.ParseIP(ip)
+		if err := config.IPAllowed(localV4); err != nil {
+			t.Fatalf("IP %s Localhost Should be accepted for WAN", ip)
+		}
+	}
+}
+
+// parseCIDRs return a possible empty list of all Network that have been parsed
+// In case of error, it returns succesfully parsed CIDRs and the last error found
```
rboyer:
```suggestion
// In case of error, it returns successfully parsed CIDRs and the last error found
```

i0rek:
@pierresouchay The only thing thats left, seem to be

> [...] some test showcasing the feature successfully keeping awareness of a blocked node from making it into the memberlist of a node configured to block it.

pierresouchay:
@i0rek yes, I have started to implement test, but I have many things to change, including the shell script to create sub networks for Mac, I'll find some time soon to work on this

pierresouchay:
@i0rek DONE

pierresouchay:
@rboyer I added the tests as required

pierresouchay:
@rboyer @i0rek @mkeeler I rebased once again, do you think you might have a look so I could advance on https://github.com/hashicorp/consul/pull/5916 ?

mkeeler:
@pierresouchay I am giving this another look now.

config.go
pierresouchay: fixed

config.go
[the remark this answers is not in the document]
rboyer: I think you forgot to delete this copy.

config.go
```diff
@@ -290,6 +322,24 @@ func DefaultWANConfig() *Config {
 	return conf
 }
 
+// IPMustBeChecked return true if IPAllowed must be called
+func (c *Config) IPMustBeChecked() bool {
+	return c.CidrsAllowed != nil
```
rboyer, line 327:
```suggestion
	return len(c.CidrsAllowed) > 0
```
Generally every time I treat a `nil` slice differently than an empty slice I ultimately come to regret it. As it doesn't make much sense to have en empty cidr allow list mean "nothing is allowed" since that would mean memberlist couldn't function, it should be semantically fine to just check for emptyness instead of nilness.

config.go
```diff
@@ -228,6 +230,35 @@ type Config struct {
 	// RequireNodeNames controls if the name of a node is required when sending
 	// a message to that node.
 	RequireNodeNames bool
+	// If nil, allow any connection (default), otherwise specify all networks
+	// allowed to connect (you must specify IPv6/IPv4 separately)
+	// Using [] will block all connections.
+	CidrsAllowed []net.IPNet
```
rboyer, line 236: trivial: Just realized as this is an initialism, go conventions say it should be `CIDRsAllowed` or `AllowedCIDRs` like `net.ParseCIDR` and `net.CIDRMask`.

net.go
```diff
@@ -635,6 +657,11 @@ func (m *Memberlist) handleAlive(buf []byte, from net.Addr) {
 		live.Port = uint16(m.config.BindPort)
 	}
 
+	if err := m.ensureCanConnect(from); err != nil {
```
rboyer, line 660: Since you are using socket information (e.g. `from net.Addr`) provided in the calling arguments, I think this `ensureCanConnect` should happen as the first line in this method.

rboyer: Edge case:
There is a subtle difference between the socket info `from net.Addr` (used here) and `(*alive).Addr` (used in `aliveNode()`). The former comes straight from our side of the socket firing data into the memberlist implementation and the latter is addressing information sent _over_ that socket from the other side. In almost all cases you would hope that these are the same.

In one very new case (https://github.com/hashicorp/consul/pull/6884) these can differ. The socket address will be functionally useless as it can come from a mesh gateway envoy local to your datacenter, but the `alive{}` structure will contain the original address as the sender views themselves.

So to be absolutely correct here, you should run *both* the socket address *and* the `(*alive).Addr` address through your `ensureCanConnect` function and block it if either fails.

config.go
pierresouchay: DONE

rboyer:
OK the last comments I left should be pretty quick to fix and after that we are golden.

net.go
pierresouchay: @rboyer In `handleAlive()`, `aliveNode()` is evently called in and already contains `m.config.IPAllowed(a.Addr)` ... do you want me to add another check anyway in `handleAlive()` anyway ?

rboyer: Oh right. I guess we can skip the double check in this function given that the latter checks will cover it.

config.go
```diff
@@ -271,6 +305,7 @@ func DefaultLANConfig() *Config {
 
 		HandoffQueueDepth: 1024,
 		UDPBufferSize:     1400,
+		CidrsAllowed:      nil, // same as allow all
```
rboyer, line 308:
```suggestion
		CIDRsAllowed:      nil, // same as allow all
```

pierresouchay:
@rboyer Thank you for the review, I think I resolved all your issues

config_test.go
```diff
@@ -0,0 +1,128 @@
...
+		if err := config.IPAllowed(localV4); err != nil {
+			t.Fatalf("IP %s Localhost Should be accepted for WAN", ip)
+		}
+	}
+}
+
+// parseCIDRs return a possible empty list of all Network that have been parsed
+// In case of error, it returns successfully parsed CIDRs and the last error found
+// If nil is given it returns nil, nil
+func parseCIDRs(v []string) ([]net.IPNet, error) {
```
rboyer, line 41: Which one of `parseCIDRs` and `ParseCIDRs` do you want to keep?

net.go
```diff
@@ -622,12 +622,45 @@ func (m *Memberlist) handleSuspect(buf []byte, from net.Addr) {
...
+		m.logger.Printf("[DEBUG] memberlist: Blocked alive message: %s %s", err, LogAddress(from))
+		return
+	}
 	var live alive
 	if err := decode(buf, &live); err != nil {
 		m.logger.Printf("[ERR] memberlist: Failed to decode alive message: %s %s", err, LogAddress(from))
 		return
 	}
+	innerIP := net.IP(live.Addr)
```
rboyer, line 657: I think this block needs to be wrapped with `if m.config.IPMustBeChecked() { ... }` to avoid the net.IP cast when not configured.

config_test.go
pierresouchay: I kept the public one since it might be useful for the program using the lib

pierresouchay:
hu... fun, unit tests do not work anymore :(

pierresouchay:
@rboyer  Ok, I found the issue with the unit tests, now, Join() by default assert that Name / Addr is set by default, so my unit tests were not working anymore, but last commit fixes that.

On another topic, the change you asked to use:
```
func (c *Config) IPMustBeChecked() bool {
	return len(c.CIDRsAllowed) > 0
}
```
instead of cheking nil also broke another test (I fixed it), that was asserting that AllowedCIDRs = [] was blocking everything.

When I wrote it initially, I thought in terms of security that blocking everything is always a nice feature, but for the project itself, it might defeat its own purpose, so I don't think it is a big deal.

An finally, the unit tests are not run in the PR, not a big deal, but it might be a good addition

I think everything is ok now

rboyer:
We can tag a new version of this after you start using this from serf or consul, in case you discover a need to do any last minute refactor of this feature while attempting to use it.

pierresouchay:
@rboyer Thanks, I'll try the integration in Consul soon, but it should be straigforward

pierresouchay:
@rboyer I made the PR in Consul some time ago, maybe you could have a look at  https://github.com/hashicorp/consul/pull/7628 ?

Thank you

PR #200, opened 2019-09-09
handle TCP transports properly

i0rek:
If there would be a way to detect if a transport is TCP or UDP I would prefer that over implementing `ProtocolType()`.

The underlying issue is that UDP and TCP are different. And since any transport is allowed, memberlist must account for both cases. For example with TCP a ping fails while sending, while with UDP there is no response.

### DIFF
--- a/state.go
+++ b/state.go
@@ -272,6 +272,12 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	// soon as possible.
 	deadline := sent.Add(probeInterval)
 	addr := node.Address()
+
+	// Arrange for our self-awareness to get updated.
+	var awarenessDelta int
+	defer func() {
+		m.awareness.ApplyDelta(awarenessDelta)
+	}()
 	if node.State == stateAlive {
 		if err := m.encodeAndSendMsg(addr, pingMsg, &ping); err != nil {
 			m.logger.Printf("[ERR] memberlist: Failed to send ping: %s", err)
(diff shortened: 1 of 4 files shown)


state.go
```diff
@@ -272,6 +272,12 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	// soon as possible.
 	deadline := sent.Add(probeInterval)
 	addr := node.Address()
+
+	// Arrange for our self-awareness to get updated.
+	var awarenessDelta int
+	defer func() {
+		m.awareness.ApplyDelta(awarenessDelta)
+	}()
```
i0rek: This is moved up from below, because we might've jump over the place where the call was previously. At this time `awarenessDelta` is `0` and nothing happens if it is applied.

state.go
```diff
@@ -281,7 +287,12 @@ func (m *Memberlist) probeNode(node *nodeState) {
 		var msgs [][]byte
 		if buf, err := encode(pingMsg, &ping); err != nil {
 			m.logger.Printf("[ERR] memberlist: Failed to encode ping message: %s", err)
-			return
+			switch m.transport.ProtocolType() {
+			case ProtocolTypeTCP:
+				goto TCPCONTINUE
+			default:
+				return
+			}
```
i0rek: When using a TCP protocol, this used to return instantly when the other end was not available. That prevent this node from being marked as suspect and failed.

state.go
```diff
@@ -336,6 +349,7 @@ func (m *Memberlist) probeNode(node *nodeState) {
 		m.logger.Printf("[DEBUG] memberlist: Failed ping: %v (timeout reached)", node.Name)
 	}
 
+TCPCONTINUE:
```
i0rek: This label jumps over the code that waits for the response to the UDP request, since we know when it didn't happen when using TCP.

state.go
```diff
@@ -360,49 +374,51 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
-			if err != nil {
-				m.logger.Printf("[ERR] memberlist: Failed fallback ping: %s", err)
-			} else {
-				fallbackCh <- didContact
-			}
-		}()
-	} else {
-		close(fallbackCh)
-	}
+	if m.transport.ProtocolType() == ProtocolTypeUDP {
```
i0rek: When already using TCP, there is no point in retrying the same request. No code changes other than the indentation.

PR #201, opened 2019-09-10
handle TCP transports properly

i0rek:
If there would be a way to detect if a transport is TCP or UDP I would prefer that over implementing `ProtocolType()`.

The underlying issue is that UDP and TCP are different. And since any transport is allowed, memberlist must account for both cases. For example with TCP a ping fails while sending, while with UDP there is no response.

### DIFF
--- a/state.go
+++ b/state.go
@@ -272,10 +272,21 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	// soon as possible.
 	deadline := sent.Add(probeInterval)
 	addr := node.Address()
+
+	// Arrange for our self-awareness to get updated.
+	var awarenessDelta int
+	defer func() {
+		m.awareness.ApplyDelta(awarenessDelta)
+	}()
 	if node.State == stateAlive {
 		if err := m.encodeAndSendMsg(addr, pingMsg, &ping); err != nil {
 			m.logger.Printf("[ERR] memberlist: Failed to send ping: %s", err)
-			return
+			switch m.transport.ProtocolType() {
+			case ProtocolTypeTCP:
+				goto TCPCONTINUE
+			default:
+				return
+			}
 		}
 	} else {
 		var msgs [][]byte
@@ -296,7 +307,12 @@ func (m *Memberlist) probeNode(node *nodeState) {
 		compound := makeCompoundMessage(msgs)
 		if err := m.rawSendMsgPacket(addr, &node.Node, compound.Bytes()); err != nil {
 			m.logger.Printf("[ERR] memberlist: Failed to send compound ping and suspect message to %s: %s", addr, err)
-			return
+			switch m.transport.ProtocolType() {
+			case ProtocolTypeTCP:
+				goto TCPCONTINUE
+			default:
+				return
+			}
 		}
 	}
 
@@ -305,10 +321,7 @@ func (m *Memberlist) probeNode(node *nodeState) {
 	// which will improve our health until we get to the failure scenarios
 	// at the end of this function, which will alter this delta variable
 	// accordingly.
-	awarenessDelta := -1
-	defer func() {
-		m.awareness.ApplyDelta(awarenessDelta)
-	}()
+	awarenessDelta = -1
 
 	// Wait for response or round-trip-time.
 	select {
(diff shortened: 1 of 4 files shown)


i0rek:
@mkeeler thanks for looking at it! I am trying to imagine what your transport would look like and I am not sure what you have in mind. I don't think the transport should handle these case, since they are not transport internal, but memberlist internal. This code changes the flow in `probeNode`, this is nothing the transport should be doing. 

The code in `probeNode` assumed a UDP protocol and I added the TCP part. That doesn't seem to bad to me. 
If I had to abstract it more, I would probably end up with a `Prober` interface and `TCPProber` and `UDPProber` implementing it.

transport.go
```diff
@@ -62,4 +69,6 @@ type Transport interface {
 	// Shutdown is called when memberlist is shutting down; this gives the
 	// transport a chance to clean up any listeners.
 	Shutdown() error
+
+	ProtocolType() ProtocolType
```
rboyer: Note this is a breaking change to the exported interface. Any custom implementations of the interface will no longer implement this interface.

An alternative would be making a new interface like:
```
type TypedTransport interface {
  Transport
  ProtocolType() ProtocolType
}
```
Then you just have to do optional type assertions to see if the transport implements the extra method. I recently did this in memberlist when i had to change `Broadcast`: https://github.com/hashicorp/memberlist/blob/master/queue.go#L128

state.go
```diff
@@ -272,10 +272,21 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
 	deadline := sent.Add(probeInterval)
 	addr := node.Address()
+
+	// Arrange for our self-awareness to get updated.
+	var awarenessDelta int
+	defer func() {
+		m.awareness.ApplyDelta(awarenessDelta)
+	}()
 	if node.State == stateAlive {
 		if err := m.encodeAndSendMsg(addr, pingMsg, &ping); err != nil {
```
rboyer: An alternative to the `ProtocolType() ProtocolType` interface change would be to have the `WriteTo()` method return a special error if the send has hard failed (in the case of TCP), maybe something like:
```
type TransportError struct {
  Err      error
  HardFailure bool // bikeshed this term; this used just for example here
}

// for go 1.13 errors.Unwrap() compatibility
func (e *TransportError) Unwrap() error { return e.Err }

// so the underlying error is what is printed if pretty printed
func (e *TransportError) Error() string { return e.Err.Error() }
```

Then (pre go 1.13) you can check:
```
err := m.encodeAndSendMsg(...)
if err != nil {
  // ... your current logging logic (the Error() impl above will imply the messages are printed identically) ...
  if terr, ok := err.(*TransportError); ok && terr.HardFailure {
    // ... your current TCP logic
  } else {
   // ... your current UDP logic
  }
}
```

rboyer: If you do this, I would say your `goto` target should probably be named something like `HANDLE_HARDFAIL` or something appropriately named.

I sorta wish there were a straightforward way to avoid the goto.

state.go
```diff
@@ -360,49 +374,51 @@ func (m *Memberlist) probeNode(node *nodeState) {
...
-			if err != nil {
-				m.logger.Printf("[ERR] memberlist: Failed fallback ping: %s", err)
-			} else {
-				fallbackCh <- didContact
-			}
-		}()
-	} else {
-		close(fallbackCh)
-	}
+	if m.transport.ProtocolType() == ProtocolTypeUDP {
```
rboyer: Ugh. So my earlier suggestion about piggybacking on the error in `WriteTo` falls apart at this line.

state.go
i0rek: I renamed the label to `HANDLE_REMOTE_FAILURE` and I am using `net.OpError` to decided what to do. That way the transport doesn't have to change.

state.go
```diff
@@ -242,6 +243,21 @@ func (m *Memberlist) probeNodeByAddr(addr string) {
 	m.probeNode(n)
 }
 
+// failedRemote checks the error and decides if it indicates a failure on the
+// other end.
+func failedRemote(err error) bool {
```
rboyer: maybe add a test of this function for sanity?

state.go
i0rek: Yeah, but my `net.OpError` checking as well. This is why I took that out again. It is just an optimization. Since TCP failed before, it will likely fail again. I am fine with that, considering how broken it is atm.

state.go
i0rek: absolutely! incoming...

state.go
```diff
@@ -242,6 +243,21 @@ func (m *Memberlist) probeNodeByAddr(addr string) {
...
+			case "dial", "read", "write":
+				return true
+			}
+		}
+	}
+	return false
+}
+
 // probeNode handles a single round of failure checking on a node.
 func (m *Memberlist) probeNode(node *nodeState) {
```
rboyer: maybe also worth adding a test that forces the mock transport to fail in the manner of a tcp failure so you can assert this entire method functions as you expect

state.go
i0rek: done

PR #203, opened 2019-10-07
ci: switch to using circleci

rboyer:


.circleci/config.yml
```diff
@@ -0,0 +1,70 @@
...
+              exit 1
+            fi
+      - run: go vet ./...
+
+  go-tests:
+    executor: go
+    steps:
+      - checkout
+      - run: mkdir -p $TEST_RESULTS
+      - run: make bin
```
alvin-huang: no `make bin` in this one. I think it's just go tests.

PR #210, opened 2019-12-04
plumbing changes necessary to allow for gossip through an intermediary network hop

rboyer:


i0rek:
ended up here a couple of times already. still in draft, right?

config.go
```diff
@@ -116,6 +116,9 @@ type Config struct {
 	// indirect UDP pings.
 	DisableTcpPings bool
 
+	// TODO
```
s-christoff: 🤔

mock_transport.go
```diff
@@ -115,7 +159,37 @@ func (t *MockTransport) StreamCh() <-chan net.Conn {
...
+	} else {
+		dest, ok = t.net.transportsByAddr[a.Addr]
+	}
+	if !ok {
+		return nil, fmt.Errorf("No route to %s", a)
+	}
+	return dest, nil
+}
+
+func getDatacenter(node string) (string, error) {
```
banks: This made me stop and think for a minute. Datacenter is not a concept that exists in memberlist alone currently. If this were in a public API method I'd think twice about introducing a Consul-ism into memberlist. It _technically_ is in that this isn't test code and might be imported by other packages although it is clearly intended for testing use.

So overall I think it's probably fine and I'm being pedantic but...

Actually is this even used? I see no references to `getDatacenter` in this PR.

rboyer: Good eye. IIRC this may have been used during the POC phase of developing the modified Transport over in memberlist before relocating it to its home in consul. Goodbye dead code.

config.go
rboyer: TODONE

transport.go
```diff
@@ -63,3 +64,64 @@ type Transport interface {
...
+func (a *Address) String() string {
+	if a.Name != "" {
+		return fmt.Sprintf("%s (%s)", a.Name, a.Addr)
+	}
+	return a.Addr
+}
+
+type IngestionAwareTransport interface {
+	Transport
+	// IngestPacket pulls a single packet off the conn and closes it.
```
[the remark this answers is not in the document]
rboyer: Changed it to `shouldClose bool` as a trailing argument.

PR #211, opened 2019-12-17
Fix the race condition in net.go where accessing node property w/o lock

zuweiliu:


dnephin:
Thank you for the PR! I know it's been over a year since you opened it, but hopefully a late response is better than none!

I think unfortunately this doesn't fully resolve the race because other fields have the same problem. I've opened a PR to fix a bunch of races and enable `go test -race`. I beleive I fixed this issue in https://github.com/hashicorp/memberlist/pull/232/commits/e3ce9856949d13a0394221c532da0027b96fdadb. I used a bit of a different approach, which was to stop using pointers and pass by value.

PR #213, opened 2020-01-30
stateLeft

i0rek:
This PR introduces another state: `stateLeft`. This is an attempt to help fix https://github.com/hashicorp/consul/issues/6897.

When memberlist processes a `dead` message, it will now check where it is coming from and if the reporting node is equal to the reported node, we know this node wants to leave. In this case we don't have to prevent new nodes from using the same ip.

If a node reports another node, the behavior stays the same.

~~**Note:** this PR adds a new state which makes this update backwards incompatible because consumer of this library need to change their code to handle `stateLeft`. While I think a new state is the most elegant way to solve this problem, we could also add a flag to `nodeState` and not change the states.~~

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -579,6 +579,10 @@ func (m *Memberlist) Leave(timeout time.Duration) error {
 			return nil
 		}
 
+		// This dead message is special, because Node and From are the
+		// same. This helps other nodes figure out that a node left
+		// intentionally. When Node equals From, other nodes know for
+		// sure this node is gone.
 		d := dead{
 			Incarnation: state.Incarnation,
 			Node:        state.Name,
--- a/memberlist.go
+++ b/memberlist.go
@@ -525,7 +525,7 @@ func (m *Memberlist) Members() []*Node {
 
 	nodes := make([]*Node, 0, len(m.nodes))
 	for _, n := range m.nodes {
-		if n.State != stateDead {
+		if !n.DeadOrLeft() {
 			nodes = append(nodes, &n.Node)
 		}
 	}
@@ -542,7 +542,7 @@ func (m *Memberlist) NumMembers() (alive int) {
 	defer m.nodeLock.RUnlock()
 
 	for _, n := range m.nodes {
-		if n.State != stateDead {
+		if !n.DeadOrLeft() {
 			alive++
 		}
 	}
@@ -582,6 +582,7 @@ func (m *Memberlist) Leave(timeout time.Duration) error {
 		d := dead{
 			Incarnation: state.Incarnation,
 			Node:        state.Name,
+			From:        state.Name,
 		}
 		m.deadNode(&d)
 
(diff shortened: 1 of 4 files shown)


state.go
```diff
@@ -1203,7 +1208,14 @@ func (m *Memberlist) deadNode(d *dead) {
 
 	// Update the state
 	state.Incarnation = d.Incarnation
-	state.State = stateDead
+
+	// If the dead message was send by the node itself, mark it is left
+	// instead of dead.
+	if d.Node == d.From {
+		state.State = stateLeft
```
i0rek: this is the magic!

i0rek:
Thank you for the review @mkcp! 

I don't have a very strong opinion on the naming. There is one reason, why I like `stateLeft` better: it makes it clear that the node left already, while `stateLeave` could also mean that the node is still leaving. I might totally be wrong though because english is not my native language. I know we have `leaveIntent` in serf eg and I didn't want to be ambiguous here.

i0rek:
Reportedly fixed the problem: https://github.com/hashicorp/consul/pull/7184#issuecomment-580488095.

state.go
```diff
@@ -963,7 +968,7 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 				time.Since(state.StateChange) > m.config.DeadNodeReclaimTime)
 
 			// Allow the address to be updated if a dead node is being replaced.
-			if state.State == stateDead && canReclaim {
+			if state.State == stateLeft || (state.State == stateDead && canReclaim) {
 				m.logger.Printf("[INFO] memberlist: Updating address for failed node %s from %v:%d to %v:%d",
```
banks: Should we change the log message in the new case? It's not really right any more as the node didn't "fail". Not a big deal but might be clearer?

memberlist.go
```diff
@@ -579,9 +579,14 @@ func (m *Memberlist) Leave(timeout time.Duration) error {
...
 		}
 
+		// This dead message is special, because Node and From are the
+		// same. This helps other nodes figure out that a node left
+		// intentionally. When Node equals From, other nodes know for
+		// sure this node is gone.
 		d := dead{
 			Incarnation: state.Incarnation,
 			Node:        state.Name,
+			From:        state.Name,
```
banks: This doesn't look like a new field as I can't see it added in this PR, but why was it not being set at all before? Is it only used in some cases? Or is it populated during send and this addition is redundant?

i0rek:
Thanks for the review! I missed the private bit and thought that the notification would include the new state as well. If thats not the case, we should be good!

memberlist.go
i0rek: It was previously only set when one node suspected another one. Hasn't been set before when a node send this message while leaving.

state.go
i0rek: updated the log to mention that the node might have left as well.

PR #216, opened 2020-04-06
Add public key encryption

evanphx:
This adds the ability for nodes to encrypt traffic between them using curve 25519 keys. The model does not use a trust relationship between the nodes based on the keys, the keys are used solely to further armor the encryption as it goes over the wire. This provides protection against people being able to intercept traffic and brute force the cluster wide encryption key.

Authorization to the cluster is provided by a new AccessKey config parameter, which is mixed into the shared secret derived from the public/private keys. This provides a similar access control mechanism as the current Secret Key mechanism, but without exposing the cluster to long term brute forcing.

The private keys used by each node are generated one node start, they are not stored on disk at all. This introduces the idea of an "encryption partition", where node A might send a message to node B with node B's old public key. B will thusly reject the message and node A will observe B as being down. This is effectively the same as if B were actually down. When node B started up, it sent it's new public key in alive messages to cluster members. Alive gossiping means that eventually A will observe B's new public key and the partition will be resolved.

### DIFF
--- a/encryption.go
+++ b/encryption.go
@@ -1,0 +1,260 @@
+package memberlist
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/subtle"
+	"encoding/base64"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"golang.org/x/crypto/curve25519"
+)
+
+// basic implemenation, cribbed from wireguard-go
+
+// Generate a new key to use for encryption. This is a convenience wrapper
+// to provide the key in the proper format for use in Memberlist config and
+// join Addresses.
+func GenerateKey() (private string, public string, err error) {
+	pk, err := NewPrivateKey()
+	if err != nil {
+		return "", "", err
+	}
+
+	return pk.HexString(), pk.Public().HexString(), nil
+}
+
+const KeySize = 32
+
+type ParseError struct {
+	Reason string
+	Input  string
+}
+
+func (p *ParseError) Error() string {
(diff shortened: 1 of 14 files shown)


encryption.go
```diff
@@ -0,0 +1,260 @@
...
+	"crypto/rand"
+	"crypto/subtle"
+	"encoding/base64"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"golang.org/x/crypto/curve25519"
```
abennett, line 14: Could you use [`crypto/ed25519`](https://golang.org/pkg/crypto/ed25519/) from the standard library for its types and key generation? It's available as of 1.13.

evanphx: No because ed25519 is only for signing, not Diffie-Hellman.

abennett: Ah! I had to google the difference between X25519 and ed25519 keys in encryption. It's described pretty well in [this blog](https://blog.filippo.io/using-ed25519-keys-for-encryption/). While it's technically possible to convert the keys back and forth, the X25519 keys use a different type of curve that permits constant time operations, which is what we would want here. Got it! 👍

evanphx: Yeah, you can but that's more about converting the data. The algorithm for performing the operations on the data is also different.

abennett: I understand that EdDSA is not ECDH. I got confused in thinking the keys were essentially the same and used the same curve. That was wrong. 😅

daneshcamp:
thnks
bug fixed

Issue #219, opened 2020-04-28
Regression introduced in v0.2.0 in (*Memberlist).SendBestEffort


agnivade:
Hi there,

This commit https://github.com/hashicorp/memberlist/commit/842cc1703d4e141a9fe096c37149652f1e1bb875 introduced a regression in the SendBestEffort method by taking the `String` representation of a Node instead of taking the `Address`.

This breaks the entire method because now rawSendMsgPacket will fail because `m.transport.WriteToAddress` fails. The following diff fixes it.

```diff
diff --git a/vendor/github.com/hashicorp/memberlist/memberlist.go b/vendor/github.com/hashicorp/memberlist/memberlist.go
index f46d466ea..af25ff810 100644
--- a/vendor/github.com/hashicorp/memberlist/memberlist.go
+++ b/vendor/github.com/hashicorp/memberlist/memberlist.go
@@ -568,7 +568,7 @@ func (m *Memberlist) SendBestEffort(to *Node, msg []byte) error {
        buf = append(buf, msg...)
 
        // Send the message
-       a := Address{Addr: to.String(), Name: to.Name}
+       a := Address{Addr: to.Address(), Name: to.Name}
        return m.rawSendMsgPacket(a, to, buf)
 }
```

I am not fully well-versed with the codebase, and that's why I am not sending a PR because this would need a unit test to lock the functionality.

Thanks.

PR #220, opened 2020-04-28
use address instead of node string representation

i0rek:
Fixes https://github.com/hashicorp/memberlist/issues/219.


rboyer:
For the record this regression was introduced in https://github.com/hashicorp/memberlist/pull/210

PR #230, opened 2021-02-16
Deprecate IngestionAwareTransport

dnephin:
This interface is never used by memberlist. The passed in transport may implement this interface, or any other interface, but memberlist does not need to be aware of it.

### DIFF
--- a/net_test.go
+++ b/net_test.go
@@ -804,7 +804,11 @@ func TestIngestPacket_ExportedFunc_EmptyMessage(t *testing.T) {
 	logger := log.New(logs, "", 0)
 	m.logger = logger
 
-	err := m.transport.IngestPacket(emptyConn, udp.LocalAddr(), time.Now(), true)
+	type ingestionAwareTransport interface {
+		IngestPacket(conn net.Conn, addr net.Addr, now time.Time, shouldClose bool) error
+	}
+
+	err := m.transport.(ingestionAwareTransport).IngestPacket(emptyConn, udp.LocalAddr(), time.Now(), true)
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "packet too short")
 }
--- a/transport.go
+++ b/transport.go
@@ -82,16 +82,18 @@ func (a *Address) String() string {
 	return a.Addr
 }
 
+// IngestionAwareTransport is not used.
+//
+// Deprecated: IngestionAwareTransport is not used and may be removed in a future
+// version. Define the interface locally instead of referencing this exported
+// interface.
 type IngestionAwareTransport interface {
-	Transport
-	// IngestPacket pulls a single packet off the conn, and only closes it if shouldClose is true.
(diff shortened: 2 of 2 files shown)


banks (changes requested):
While it might have been cleaner to never expose this, I don't think we can just remove it without breaking backwards compatibility gaurantees which is not allowed by Go modules/semantic versioning unless we release a v2.

The issue is that even if memberlist doesn't use this outside test code, downstream code might explicitly type switch on the `memberlist.IngestionAwareTransport` interface type or use it as a struct field or argument type [like we do in Consul](https://github.com/hashicorp/consul/blob/a4327305d1ca583ae459f1d0e3eeacc6805b4316/agent/consul/server.go#L254).

For any such consumer this would be a breaking change. I don't think it's important enough to warrant a v2 switch or breaking the rules of semantic versioning to undo this even if it could have been left out of the interface originally.

dnephin:
I agree with most of that argument, but I think a reasonable argument can be made that this is not in fact a breaking change.

It is no surprise that Consul imports it, because this was added for Consul. There is no logical reason for any other code to import this symbol because it is only used by Consul. Any other user would chose a different name for this interface and the methods and would have no reason to use this un-referenced interface from memberlist.

At the very least removing the shim and the embedding of this type in `NodeAwareTransport` is a safe change. We could leave the interface itself sitting there and document that it is not used, but it seems very unlikely that anyone would reference it. A [github search](https://github.com/search?p=2&q=IngestionAwareTransport&type=Code) shows at least no public code on github uses it.

banks:
Yeah, I doubt that it would be experienced as a breaking change by anyone - certainly not many if it is.

I see it as the better option to not break backwards compatibility of a public library in any way unless there is a _really_ good reason like a CVE or major bug which I don't think this qualifies as 🤷.  To me it doesn't seem important enough either way to spend lots of time on so I'd err for the safer option personally but won't loose sleep if others disagree 😄 .

Your suggestion of doing most of this but leaving a stub interface definition with a deprecation comment seems like a reasonable path forward to me that makes it a trivial merge again. What do you think?

Open to input from others too!

net_test.go
```diff
@@ -804,7 +804,11 @@ func TestIngestPacket_ExportedFunc_EmptyMessage(t *testing.T) {
 	logger := log.New(logs, "", 0)
 	m.logger = logger
 
-	err := m.transport.IngestPacket(emptyConn, udp.LocalAddr(), time.Now(), true)
+	type ingestionAwareTransport interface {
+		IngestPacket(conn net.Conn, addr net.Addr, now time.Time, shouldClose bool) error
+	}
```
dnephin, line 809: Using a local interface here to prevent referencing the deprecated interface.

dnephin (commented):
Sounds good, I've made those changes.

I restored the interface, but added a deprecation comment. The shim changes should be backwards compatible.

banks (approved):
LGTM thanks!

PR #232, opened 2021-03-23
Fix data races

dnephin:
Fixes #134
Fixes #113
Closes #179

This PR fixes a number of data races and runs tests in CI using `go test -race` to catch any regressions. Also updates the version of Go used in CI to 1.15. I tried upgrading to 1.16 but there are a number of problems (for some reason the $PATH in that 1.16 image is wrong, and `go vet` reports more problems https://golang.org/doc/go1.16#vet, so I'll defer that upgrade to another PR).

Lots of detail in the commit messages, best viewed by individual commit.

The `EventDelegate` is really only a partial fix for these tests. The data  race is still there, and unfortunately it seems like the interface is a problem.

PR #233, opened 2021-03-23
Fix flaky tests

dnephin:
Branched from #232

By waiting for the condition to be true, instead of sleeping a fixed amount. Occasionally these tests would fail when it took too long to run.

Also remove the use of t.Log from a logger. This would cause tests to panic because the goroutines using t.Log can outline the main test function.


banks (approved):
Nice. This looks much safer and likely quicker too!

PR #235, opened 2021-04-13
memberlist.Logger Interface, More Flexible Logger Replacement

michaelquigley:
Implemented a `memberlist.Logger` interface, allowing for more flexible logger library replacement (we use `logrus`). Also removes a bunch of `[LEVEL] memberlist:` boilerplate from log messages.

Care was taken to handle all of the logging specializations in the test cases.


michaelquigley:
Not sure what the failing test is, but locally I ran into a number of tests that don't seem to execute reliably.

If there's a `go fmt` issue, please let me know and I'll fix it promptly.

michaelquigley:
CLA signed.

PR #239, opened 2021-07-08
Fix for compound messages containing >255 messages.

stevesg:
If ever more than 255 messages are put into a compound message, then the resulting packet will be malformed leading to errors on the receiving end such as:

```
Failed to decode ping request: codec.decoder: Only encoded map or array can be decoded into a struct. (valueType: 2)
```

This change sends multiple packets if required. The alternative approach would be to limit the number of messages that `GetBroadcasts` can return, by e.g. passing in a maximum number of messages, but this would require changing the `Delegate` interface, possibly breaking downstream code.

### DIFF
--- a/memberlist_test.go
+++ b/memberlist_test.go
@@ -1200,9 +1200,9 @@ func TestMemberlist_UserData(t *testing.T) {
 
 	bindPort := m1.config.BindPort
 
-	bcasts := [][]byte{
-		[]byte("test"),
-		[]byte("foobar"),
+	bcasts := make([][]byte, 256)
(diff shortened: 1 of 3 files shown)


pracucci (commented):
This fix makes sense to me 👍

pracucci:
> This looks reasonable at first glance. I'm very curious how you found this. Can you give any details on the situation?

We're experiencing this when running large [Cortex](https://github.com/cortexproject/cortex) clusters and we found it after some debugging.

markan (approved):
This looks reasonable to me.

PR #242, opened 2021-08-24
Fix udp writes not causing nodes to become suspect

Austinpayne:
failedRemote only checks for TCP failures but probeNode uses
UDP meaning that if we get a "network is unreachable" during
a UDP write the node is never marked as suspect.

This failure mode can be reproduced by bringing down an
interface in which case the node will show its neighbors
as alive while all other nodes will properly detect the
node as failed.

For reproduction steps see: https://gist.github.com/Austinpayne/defc5178bc2cc5c29ff5ada3dc4b1260
For verifying the fix see: https://gist.github.com/Austinpayne/151876ba8841be46484ce71664ced6bc

### DIFF
--- a/state.go
+++ b/state.go
@@ -274,6 +274,11 @@ func failedRemote(err error) bool {
 			case "dial", "read", "write":
 				return true
 			}
+		} else if strings.HasPrefix(t.Net, "udp") {
+			switch t.Op {
+			case "write":
+				return true
+			}
 		}
 	}
 	return false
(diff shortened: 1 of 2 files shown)


Austinpayne:
Fixes https://github.com/hashicorp/serf/issues/627

cc: @i0rek @rboyer @mkeeler

kyhavlov (approved):
Hi @Austinpayne, this looks great, thanks for the PR and the clean reproduction setup!

Austinpayne:
Thanks @kyhavlov. Should I target a PR for serf as well?

edsharp:
Awesome - I've made a local build of `serf` against 7e2219b23f33 and it fixes the issue for me :)

Can I help with tagging `memberlist` and tagging a new `serf`?

PR #248, opened 2021-11-05
Fix a data race in suspicion Confirm

lodthe:
Maps are not thread-safe, so data access must be synchronized.

### DIFF
--- a/suspicion.go
+++ b/suspicion.go
@@ -2,6 +2,7 @@ package memberlist
 
 import (
 	"math"
+	"sync"
 	"sync/atomic"
 	"time"
 )
@@ -39,7 +40,8 @@ type suspicion struct {
 
 	// confirmations is a map of "from" nodes that have confirmed a given
 	// node is suspect. This prevents double counting.
-	confirmations map[string]struct{}
+	confirmations   map[string]struct{}
+	confirmationsMu sync.Mutex
 }
 
 // newSuspicion returns a timer started with the max time, and that will drive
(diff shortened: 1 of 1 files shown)


eculver:
I'm curious how you discovered this problem given that we are currently running tests with the `--race` flag? Would you mind sharing some of your process/context here?

lodthe:
> I'm curious how you discovered this problem given that we are currently running tests with the `--race` flag? Would you mind sharing some of your process/context here?
I didn't use memberlist somewhere, I was just exploring the implementation of the algorithm. 

It just caught my eye that some modifications are atomic (e.g., when modifying the number of confirmations), but the map modifications are not synchronized. In my mind, modifications of both of these variables must be either synchronized or not.

dnephin (commented):
Thank you for your interest in improving memberlist!

From my reading of the code, the only place this `suspicion.Confirm` method is called is from here:

https://github.com/hashicorp/memberlist/blob/923f1b205dd4653f2ea35e1c9531088e52053aa0/state.go#L1140-L1159

At the top of that function the `nodeLock` is held. I think that means that in practice this is not a data race. There is a lock, but it is being managed by the caller to this function.

I think it might be valuable to document this assumption in the code, to prevent changes that might break the assumption, but I don't think we want to introduce a lock for this map if it's never accessed concurrently.

dnephin:
Thanks again for the PR! We appreciate the desire to improve memberlist. Since it seems this is safe in practice we are going to close this PR.

PR #252, opened 2021-12-23
Add warning message for remote node state limit

bwaters:
Warn before the memberlist starts to fail because of unchangable limit

### DIFF
--- a/net.go
+++ b/net.go
@@ -1089,6 +1089,11 @@ func (m *Memberlist) decryptRemoteState(bufConn io.Reader, streamLabel string) (
 	moreBytes := binary.BigEndian.Uint32(cipherText.Bytes()[1:5])
 	if moreBytes > maxPushStateBytes {
 		return nil, fmt.Errorf("Remote node state is larger than limit (%d)", moreBytes)
+
+	}
+	//Start reporting the size before you cross the limit
+	if moreBytes > .6 * maxPushStateBytes {
+		m.logger.Printf("[WARN] memberlist: Remote node state size is %d approaching limit (%d)",moreBytes,maxPushStateBytes )
 	}
 
 	// Read in the rest of the payload


bwaters:
We did some debugging to figure out why we hit the limit and uncovered another bug.  We had a huge number of old nodes in the StateLeft state.

net.go
```diff
@@ -1089,6 +1090,12 @@ func (m *Memberlist) decryptRemoteState(bufConn io.Reader, streamLabel string) (
 	moreBytes := binary.BigEndian.Uint32(cipherText.Bytes()[1:5])
 	if moreBytes > maxPushStateBytes {
 		return nil, fmt.Errorf("Remote node state is larger than limit (%d)", moreBytes)
+
+	}
+
+	//Start reporting the size before you cross the limit
+	if moreBytes > uint32(math.Floor(.06*maxPushStateBytes)) {
```
kisunji, line 1097: Confirming 6% was intentional? Should it be a higher threshold like 80 or 90?

net.go
```diff
@@ -1089,6 +1090,12 @@ func (m *Memberlist) decryptRemoteState(bufConn io.Reader, streamLabel string) (
...
+		m.logger.Printf("[INFO] memberlist: Remote node state size is %d  limit is (%d)", moreBytes, maxPushStateBytes)
```
FFMMM, line 1098: [nit]

```suggestion
		m.logger.Printf("[INFO] memberlist: Remote node state size is (%d) limit is (%d)", moreBytes, maxPushStateBytes)
```

net.go
bwaters: Changed for debugging.  Intention was to have it at 60% to give time to upgrade fix things before you cross the mark.  If you do hit this bug chances are your cluster is large enough that upgrading is non-trivial.

Issue #253, opened 2022-01-05
StateLeft members are never reaped


bwaters:
Unlike StateDead members old nodes that gracefully leave are never reaped.  They just permanently increase the node list and sit inside the push/pull state.  

Eventually on a large cluster with lots of churn you will hit maxPushStateBytes and push pull stops working completely

bwaters:
Looks like this bug was introduced when StateLeft was implemented.
https://github.com/hashicorp/memberlist/commit/237d410aa2bf83254678ef78dd638480780e54a2

Looks like most places in the code where == StateDead were updated to n.DeadOrLeft() but this one was missed.

bwaters:
Once a sufficient number of nodes reach the maxPushStateBytes limit then new members are unable to join the consul cluster

PR #254, opened 2022-01-05
Purge left nodes along with dead ones (#253)

bwaters:
Keep the state from infinitely expanding

### DIFF
--- a/util.go
+++ b/util.go
@@ -102,7 +102,7 @@ func moveDeadNodes(nodes []*nodeState, gossipToTheDeadTime time.Duration) int {
 	numDead := 0
 	n := len(nodes)
 	for i := 0; i < n-numDead; i++ {
-		if nodes[i].State != StateDead && nodes[i].State != StateLeft {
+		if !nodes[i].DeadOrLeft() {
 			continue
 		}
 
--- a/util.go
+++ b/util.go
@@ -96,7 +96,7 @@ func pushPullScale(interval time.Duration, n int) time.Duration {
 	return time.Duration(multiplier) * interval
 }
 
-// moveDeadNodes moves nodes that are dead and left and beyond the gossip to the dead interval
+// moveDeadNodes moves dead and left nodes that that have not changed during the gossipToTheDeadTime interval
 // to the end of the slice and returns the index of the first moved node.
 func moveDeadNodes(nodes []*nodeState, gossipToTheDeadTime time.Duration) int {
 	numDead := 0
--- a/util.go
+++ b/util.go
@@ -96,13 +96,13 @@ func pushPullScale(interval time.Duration, n int) time.Duration {
 	return time.Duration(multiplier) * interval
 }
 
-// moveDeadNodes moves nodes that are dead and beyond the gossip to the dead interval
+// moveDeadNodes moves nodes that are dead and left and beyond the gossip to the dead interval
 // to the end of the slice and returns the index of the first moved node.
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -96,13 +96,13 @@ func pushPullScale(interval time.Duration, n int) time.Duration {
...
 
-// moveDeadNodes moves nodes that are dead and beyond the gossip to the dead interval
+// moveDeadNodes moves nodes that are dead and left and beyond the gossip to the dead interval
 // to the end of the slice and returns the index of the first moved node.
 func moveDeadNodes(nodes []*nodeState, gossipToTheDeadTime time.Duration) int {
 	numDead := 0
 	n := len(nodes)
 	for i := 0; i < n-numDead; i++ {
-		if nodes[i].State != StateDead {
+		if nodes[i].State != StateDead && nodes[i].State != StateLeft {
```
i0rek, line 105:
```suggestion
		if !nodes[i].DeadOrLeft() {
```

i0rek: This is a bit easier to read in my mind and aligns with the rest of the code.

i0rek (changes requested):
This looks good to me, thanks for the detailed explanation in the corresponding issue. Could you modify `TestMoveDeadNodes` to have a nodestate in `StateDead` to make sure it is working as intended?

util.go
```diff
@@ -96,13 +96,13 @@ func pushPullScale(interval time.Duration, n int) time.Duration {
 	return time.Duration(multiplier) * interval
 }
 
-// moveDeadNodes moves nodes that are dead and beyond the gossip to the dead interval
+// moveDeadNodes moves nodes that are dead and left and beyond the gossip to the dead interval
```
dnephin, line 99: I'm not sure I follow this godoc comment, is there an extra word or a missing word here?

Does this sound right ?

```suggestion
// moveDeadNodes moves dead and left nodes that that have not changed during the gossipToTheDeadTime duration
```

dnephin (commented):
Thanks for the fix!  One suggestion for the godoc.

Is there any way we can test this from within memberlist? If a full integration test makes it difficult to reproduce, maybe just a test of `moveDeadNodes`, to show that it moves nodes in both states? Edit: Oh, I missed that Hans already suggested a test for this, that sounds good!

util.go
markan: As far as a test goes, maybe something in state_test:TestMemberList_ResetNodes?

util_test.go
```diff
@@ -204,14 +218,19 @@ func TestMoveDeadNodes(t *testing.T) {
 			if nodes[i].State != StateDead {
 				t.Fatalf("Bad state %d", i)
 			}
+		case 3:
+			//Recently left node should remain at 3
+			if nodes[i].State != StateLeft {
+				t.Fatalf("Bad State #{i}")
```
kisunji, line 224:
```suggestion
				t.Fatalf("Bad State %d", i)
```

bwaters: If we are changing it here we should change it in the rest of the test.

kisunji: Perhaps you are using Goland with the default code folding? It compacts format strings like
```go
t.Fatalf("Bad State %d", i)
```
into
```go
t.Fatalf("Bad State #{i}")
```
But the latter is not a valid format syntax

bwaters: yes you were right ide was playing tricks on me

i0rek (approved):
LGTM

PR #256, opened 2022-01-27
clean code

ls-2018:
Signed-off-by: ls-2018 <acejilam@gmail.com>

Make code cleaner

### DIFF
--- a/net_transport.go
+++ b/net_transport.go
@@ -3,14 +3,14 @@ package memberlist
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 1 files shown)


dnephin (commented):
Thank you for the PR! It seems like this changes the behaviour, and from the PR description it sounds like you did not intend to change the behaviour.

Previously it was attempting to set a read buffer size. If the set failed it was reducing the size by 50% and trying again. The new version seems to only try to set the read buffer once.

Did you want to change the behaviour? Could you tell us more about the motivation for this change?

Issue #234, opened 2021-03-30
More user-friendly error message for connectivity issues


krastin:
Hello,

After a couple of cases where end-users were perplexed by the following error message, I thought it might be better to state that we were able to connect over TCP explicitly, instead of referring to just 'connection', so that we can make the customer's troubleshooting easier?

https://github.com/hashicorp/memberlist/blob/a43ca4a5f16cd26cbf7b27eb613982bfd09c1753/state.go#L472

Something in the lines of "memberlist: Was able to establish TCP session to %s but other UDP probes failed"? Would that make sense? It will certainly help the user to know what exactly to follow-up on.

Thanks,
Krastin

acnodal-tc:
The https://gitlab.com/purelb/purelb project uses memberlist and one of our users is seeing this message. Do you have any advice on how to debug the cause of it? It sounds like this message indicates that UDP packets from one node to another are being dropped? Is this correct?

krastin:
As far as I understand the code myself, you are correct

PR #258, opened 2022-02-09
Explicit transport protocols in state errors: TCP or UDP

krastin:
As described in issue #234 , when troubleshooting connectivity issues, end-users routinely get confused that Memberlist **ping** messages are ICMP packets.

I'm basing my edits on the fact that Memberlist pings were already referred to as `UDP Ping` here:

https://github.com/hashicorp/memberlist/blob/a43ca4a5f16cd26cbf7b27eb613982bfd09c1753/state.go#L535

Anytime there is a mention of `ping`, I've added explicitly whether this is an UDP or TCP segment.


kisunji (approved):
Thank you for this change! LGTM

PR #270, opened 2022-08-02
Add metrics labels to all metrics

huikang:
Cherry pick from the [metrics-label-backport branch](https://github.com/hashicorp/memberlist/commits/metrics-labels-backport)

### DIFF
--- a/config.go
+++ b/config.go
@@ -244,10 +244,14 @@ type Config struct {
 	// RequireNodeNames controls if the name of a node is required when sending
 	// a message to that node.
 	RequireNodeNames bool
+
 	// CIDRsAllowed If nil, allow any connection (default), otherwise specify all networks
 	// allowed to connect (you must specify IPv6/IPv4 separately)
 	// Using [] will block all connections.
 	CIDRsAllowed []net.IPNet
+
+	// MetricLabels is a map of optional labels to apply to all metrics emitted.
+	MetricLabels map[string]string
 }
 
 // ParseCIDRs return a possible empty list of all Network that have been parsed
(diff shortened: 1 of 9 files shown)


config.go
```diff
@@ -244,10 +244,14 @@ type Config struct {
...
 	// a message to that node.
 	RequireNodeNames bool
+
 	// CIDRsAllowed If nil, allow any connection (default), otherwise specify all networks
 	// allowed to connect (you must specify IPv6/IPv4 separately)
 	// Using [] will block all connections.
 	CIDRsAllowed []net.IPNet
+
+	// MetricLabels is a map of optional labels to apply to all metrics emitted.
+	MetricLabels map[string]string
```
rboyer, line 254: Would it make sense to match the serf config and have this be a slice of metric labels?

huikang: Either way looks good to me. If we switch to `MetricLabels []metrics.Label`, we can get rid of `mapToLabels` and pass []metrics.Label all the way from consul to memberlist, right?

rboyer: yep

memberlist.go
```diff
@@ -198,6 +203,8 @@ func newMemberlist(conf *Config) (*Memberlist, error) {
 		}
 	}
 
+	metricLabels := conf.MetricLabels
```
rboyer, line 206: nit: why make a local variable for two uses below?

rboyer (approved):

memberlist.go
huikang: oops, I forgot to delete the variables after change the type; we used to have `metricLabels := mapToLabels(conf.MetricLabels)`

Fixed now.

PR #275, opened 2022-08-19
Adding the regular emitting of the memberlist broadcast queue

jmurret:
## Implementation
This implementation follows the pattern that exists in [serf](https://github.com/hashicorp/serf/blob/830be12c38e509faa64a2a650629e91c8930a040/serf/serf.go#L432) for capturing intent queue metrics.  

## Verification
I verified this change by following the [Layer 7 Observability Learn Guide](https://learn.hashicorp.com/tutorials/consul/kubernetes-layer7-observability) to run prometheus and Grafana and view the metrics in a Grafana Dashboard.

### Before
Below you can see that `consul_memberlist_queue_Intent` (this has since been renamed to `consul_memberlist_queue_broadcast`) is not available as a metric.

Screen Shot 2022-08-25 at 4 40 44 PM

### After
Below you can see that `consul_memberlist_queue_Intent` (this has since been renamed to `consul_memberlist_queue_broadcast`) is available as a metric and is showing data in the dashboard.
Screen Shot 2022-08-25 at 4 33 37 PM

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -233,7 +233,7 @@ func newMemberlist(conf *Config) (*Memberlist, error) {
 	go m.streamListen()
 	go m.packetListen()
 	go m.packetHandler()
-	go m.checkQueueDepth("Intent", m.broadcasts)
+	go m.checkBroadcastQueueDepth()
 	return m, nil
 }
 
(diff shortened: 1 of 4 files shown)


memberlist.go
```diff
@@ -233,6 +233,7 @@ func newMemberlist(conf *Config) (*Memberlist, error) {
 	go m.streamListen()
 	go m.packetListen()
 	go m.packetHandler()
+	go m.checkQueueDepth("Intent", m.broadcasts)
```
huikang, line 236: - Shall we use all lower-case for the name like some other places?

https://github.com/hashicorp/memberlist/blob/e6ff9b2d87a3f0f3f04abb5672ada3ac2a640223/state.go#L295-L297

- Also I'm not sure if `Intent` is a good name here. "Intent" is consistent with the name in serf, but make it difficult to infer in the memberlist library. If we choose a different name like "broadcast" or "transmit", then we may need to update the consul telemetry page about the correlation between memberlist.broadcase and serf.Intent.

memberlist.go
```diff
@@ -776,3 +777,17 @@ func (m *Memberlist) changeNode(addr string, f func(*nodeState)) {
 	n := m.nodeMap[addr]
 	f(n)
 }
+
+// checkQueueDepth periodically checks the size of a queue to see if
+// it is too large
+func (m *Memberlist) checkQueueDepth(name string, queue *TransmitLimitedQueue) {
```
huikang, line 783: I'm not sure if we need a name and queue as parameters in the method. Since memberlist struct only has one queue.

huikang (approved):
LGTM.

Added some non-blocking comments; I will leave it to RB to decide.

memberlist.go
jmurret: Great callout.  I've changed the name and signature to `checkBroadcastQueueDepth()` and within the function it passes `broadcast` as the metric name.  Thank you for calling this out.

memberlist.go
jmurret: See above comment this has been resolved.  Thank you for calling this out.

huikang: Looks great. Thanks for the update.

rboyer (approved):

PR #276, opened 2022-08-26
Emit metrics for local memberlist size and remote memberlist size

jmurret:
## Changes proposed
- emit a `"memberlist", "node", "instances"` Gauge and pass label so a gauge can be broken out by node state
  - it appears gauges are reported with the host/pod name in the metric.  so the above gauge will emit metrics in the sink (grafana) for 1 server and 3 clients as:
    - consul_server_0_memberlist_node_instances
    - consul_client_1_memberlist_node_instances
    - consul_client_2_memberlist_node_instances
    - consul_client_3_memberlist_node_instances
 
Screen Shot 2022-08-30 at 11 53 49 AM

Screen Shot 2022-08-30 at 11 53 26 AM

- emit `"memberlist", "size", "local"` Gauge that reports the size in bytes of the memberlist size before it is sent to another gossip recipient.
  - same condition as above for multiple metrics being create by hostname

Screen Shot 2022-08-30 at 11 56 46 AM
- emit `"memberlist", "size", "remote"` Gauge that reports the size of incoming memberlists from other gossip senders
  - is this helpful?  I feel like we would have to add a label for who is sending it?

### DIFF
--- a/net.go
+++ b/net.go
@@ -1042,6 +1042,9 @@ func (m *Memberlist) sendLocalState(conn net.Conn, join bool, streamLabel string
 		}
 	}
 
+	moreBytes := binary.BigEndian.Uint32(bufConn.Bytes()[1:5])
+	metrics.AddSampleWithLabels([]string{"memberlist", "size", "local"}, float32(moreBytes), m.metricLabels)
+
 	// Get the send buffer
 	return m.rawSendMsgStream(conn, bufConn.Bytes(), streamLabel)
 }
@@ -1088,6 +1091,8 @@ func (m *Memberlist) decryptRemoteState(bufConn io.Reader, streamLabel string) (
 	// Ensure we aren't asked to download too much. This is to guard against
 	// an attack vector where a huge amount of state is sent
 	moreBytes := binary.BigEndian.Uint32(cipherText.Bytes()[1:5])
+	metrics.AddSampleWithLabels([]string{"memberlist", "size", "remote"}, float32(moreBytes), m.metricLabels)
+
 	if moreBytes > maxPushStateBytes {
 		return nil, fmt.Errorf("Remote node state is larger than limit (%d)", moreBytes)
 
--- a/net.go
+++ b/net.go
@@ -1043,7 +1043,7 @@ func (m *Memberlist) sendLocalState(conn net.Conn, join bool, streamLabel string
 	}
 
 	moreBytes := binary.BigEndian.Uint32(bufConn.Bytes()[1:5])
-	metrics.AddSampleWithLabels([]string{"memberlist", "size", "local"}, float32(moreBytes), m.metricLabels)
+	metrics.SetGaugeWithLabels([]string{"memberlist", "size", "local"}, float32(moreBytes), m.metricLabels)
 
 	// Get the send buffer
 	return m.rawSendMsgStream(conn, bufConn.Bytes(), streamLabel)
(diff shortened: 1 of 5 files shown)


net.go
```diff
@@ -87,6 +87,13 @@ const (
 	maxPushPullRequests    = 128 // Maximum number of concurrent push/pull requests
 )
 
+const (
+	nodeStateAlive   = "alive"
+	nodeStateDead    = "dead"
+	nodeStateLeft    = "left"
+	nodeStateSuspect = "suspect"
```
rboyer, line 94: You could add a method on `NodeStateType` instead:

```
func (t NodeStateType) metricsString() string {
		switch n.State {
		case StateAlive:
			return "alive"
		case StateDead:
			return "dead"
		case StateSuspect:
			return "suspect"
		case StateLeft:
			return "left"
		default:
			return ""
		}
}
```

net.go
```diff
@@ -1088,6 +1123,8 @@ func (m *Memberlist) decryptRemoteState(bufConn io.Reader, streamLabel string) (
 	// Ensure we aren't asked to download too much. This is to guard against
 	// an attack vector where a huge amount of state is sent
 	moreBytes := binary.BigEndian.Uint32(cipherText.Bytes()[1:5])
+	metrics.SetGaugeWithLabels([]string{"memberlist", "size", "remote"}, float32(moreBytes), m.metricLabels)
```
jmurret, line 1126: Is this actually helpful in a gauge?  Should I remove this?  I added this here because this is where warnings and errors happen related to maxPushStateBytes.  The receiver warns that the incoming message is too large.  
Since it appears that Gauges get reported with the hostname embedded in the metric, IE `consul_server_0_memberlist_size_remote`, this metric would seem to indicate "What is the size of the last message that Server 0 received?"  if we wanted to see it broken out by sender, we could add a label, but this seems like it is not that effective and the local state is really a much better thing to follow.  Thoughts?

net.go
jmurret: great call.

net.go
```diff
@@ -1042,6 +1074,9 @@ func (m *Memberlist) sendLocalState(conn net.Conn, join bool, streamLabel string
 		}
 	}
 
+	moreBytes := binary.BigEndian.Uint32(bufConn.Bytes()[1:5])
+	metrics.SetGaugeWithLabels([]string{"memberlist", "size", "local"}, float32(moreBytes), m.metricLabels)
```
jmurret, line 1078: suggestions for a better metric name?  (also, see question below about removing the remote metric, which might make naming this easier.)

net.go
jmurret: fixed.

state.go
```diff
@@ -15,6 +15,21 @@ import (
...
+	case StateAlive:
+		return "alive"
+	case StateDead:
+		return "dead"
+	case StateSuspect:
+		return "suspect"
+	case StateLeft:
+		return "left"
+	default:
+		return ""
```
rboyer, line 29: It might make sense to return something like:
```
// Should not happen, sign that you should add a new case above.
return fmt.Sprintf("unhandled-value-%d", t)
```

jmurret: good call.  updated.

net.go
rboyer: Yeah it seems like if you keep this one (remote), it would make more sense as like a histogram so you could avoid collapsing the data points in the manner of a gauge.

jmurret: 👍 thanks @rboyer. I've made the suggested change.

rboyer (approved):
Once the library versioning bumps are done these can be documented on the consul telemetry page on the website.

PR #277, opened 2022-09-30
[DO-NOT-MERGE] Ensure messages in gossip function are the same for each node

jmurret:
This refactor addresses:
- removing the getBroadcasts() call out of the message/node loop so that we avoid the potential case that one node gets a different set of messages than another. (and therefore infection rate is diminished because for a given message because not all 3 nodes are relaying it).
- making sure that compound messages are not rebuilt for each node and we ensure they each get the same compound messages (and also less processing).

### DIFF
--- a/queue.go
+++ b/queue.go
@@ -131,13 +131,13 @@ type Broadcast interface {
 // You shoud ensure that Invalidates() checks the same uniqueness as the
 // example below:
 //
-// func (b *foo) Invalidates(other Broadcast) bool {
-// 	nb, ok := other.(NamedBroadcast)
-// 	if !ok {
-// 		return false
-// 	}
-// 	return b.Name() == nb.Name()
-// }
+//	func (b *foo) Invalidates(other Broadcast) bool {
+//		nb, ok := other.(NamedBroadcast)
+//		if !ok {
+//			return false
+//		}
+//		return b.Name() == nb.Name()
+//	}
 //
 // Invalidates() isn't currently used for NamedBroadcasts, but that may change
 // in the future.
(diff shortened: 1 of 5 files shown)


rboyer:
LGTM, assuming you don't merge it.

queue.go
```diff
@@ -320,7 +320,7 @@ func (q *TransmitLimitedQueue) GetBroadcasts(overhead, limit int) [][]byte {
 			id:        math.MaxInt64,
 		}
 		lessThan := &limitedBroadcast{
-			transmits: transmits + 1,
+			transmits: transmits + transmissionCount,
```
jmurret, line 323: modify b-tree search to consider k nodes when traversing levels.

queue.go
```diff
@@ -352,15 +352,15 @@ func (q *TransmitLimitedQueue) GetBroadcasts(overhead, limit int) [][]byte {
...
+		if keep.transmits+transmissionCount >= transmitLimit {
 			keep.b.Finished()
 		} else {
 			// We need to bump this item down to another transmit tier, but
 			// because it would be in the same direction that we're walking the
 			// tiers, we will have to delay the reinsertion until we are
 			// finished our search. Otherwise we'll possibly re-add the message
 			// when we ascend to the next tier.
-			keep.transmits++
+			keep.transmits += transmissionCount
```
jmurret, line 363: increment transmits based on k nodes rather than incrementing 1.

state.go
```diff
@@ -616,31 +616,34 @@ func (m *Memberlist) gossip() {
 		bytesAvail -= encryptOverhead(m.encryptionVersion())
 	}
 
-	for _, node := range kNodes {
-		// Get any pending broadcasts
-		msgs := m.getBroadcasts(compoundOverhead, bytesAvail)
-		if len(msgs) == 0 {
-			return
+	msgs := m.getBroadcasts(compoundOverhead, bytesAvail, len(kNodes))
```
jmurret, line 619: In this file:
- move get broadcasts outside of loop
- build messages once and ensure the same messages (and improved processing) gets sent to each node.

PR #288, opened 2023-07-20
Fix go-metrics dep

kmulvey:
it seems to have moved from github.com/armon/go-metrics to github.com/hashicorp/go-metrics


kmulvey:
sorry, someone else got it first

PR #292, opened 2023-10-19
Upgrade to go-msgpack/v2 2.1.1

swenson:
This adds the option to specify what version of the `time.Time` encoding format is desired. The default option is `false`, which preserves compatibility with the current dependency of
`hashicorp/go-msgpack 0.5.3` in `go.mod`, but users of this library will probably want to override the the setting.

While we're here, test against the latest versions of Go and upgrade the `go.mod` file.

I tested that this appears to work with Nomad.


.github/workflows/check.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        GO_VERSION: [ "1.16","1.17","1.18" ]
+        GO_VERSION: [ "1.16","1.17","1.18","1.19","1.20","1.21"" ]
```
dhiaayachi, line 18: Shouldn't we only stick with the supported versions of go 🤔 ?

config.go
```diff
@@ -260,6 +260,12 @@ type Config struct {
 	// QueueCheckInterval is the interval at which we check the message
 	// queue to apply the warning and max depth.
 	QueueCheckInterval time.Duration
+
+	// MsgpackUseNewTimeFormat when set to true, force the underlying msgpack
+	// codec to use the new format of time.Time when encoding (used in
+	// go-msgpack v1.1.5 by default). Decoding is not affected, as all
+	// go-msgpack v2.1.0+ decoders know how to decode both formats.
+	MsgpackUseNewTimeFormat bool
```
dhiaayachi, line 268: I'm not sure I fully understand, shouldn't this comment be: 
 ```suggestion
	// MsgpackUseNewTimeFormat when set to false, force the underlying msgpack
	// codec to use the old format of time.Time when encoding (used in
	// go-msgpack v1.1.5 by default). Decoding is not affected, as all
	// go-msgpack v2.1.0+ decoders know how to decode both formats.
	MsgpackUseNewTimeFormat bool
```

dhiaayachi (approved):
Thank you for leading this effort @swenson! 

I left few comments but nothing blocking.

.github/workflows/check.yml
swenson: I only added versions. I didn't want to remove anything. :)

config.go
swenson: I believe those statements are equivalent.

The default is `false` though, so I went with documenting what happens when you change the value to true. (I have nearly the same comment everywhere else, with the same parameter name.)

swenson:
(test failures appear to be unrelated to these changes)

swenson:
Thanks!

Issue #293, opened 2023-11-09
Checking the state of the Node in NotifyLeave


krivoguzovVlad:
I would like to check whether the service was stopped gracefully in `NotifyLeave`. The callback accepts a `Node` with a `State` field, but it is always equal to `StateAlive`.

Having looked at the code, I believe that the problem is that the `State` field is simultaneously declared in both `Node` and `nodeState` structs, while `Node` is nested in `nodeState`. Could someone explain if there are reasons to do this?

I see several solutions, but unfortunately they are not backwards compatible:
- add a getter and setter for `nodeState`, which will also change `Node.State`. At the same time, make the `State` field private.
- remove `State` field from `nodeState`
- remove the `State` field from `Node`, and add a separate parameter with the state to `NotifyLeave`

Issue #266, opened 2022-07-27
Feature request: Update state of node in memberlist


y-eight:
Currently, the state property of a node in the memberlist is not updated yet.

```
for _, member := range mesh.Members() {
	fmt.Printf("Member: %s %v\n", member.Name, member.State)
}
```
2 nodes in cluster:
```
Member: node8082 0
Member: node8081 0
```
2nd Node timeout - not reachable anymore. Global health score changes to a higher score.
The state of the 2nd Node in the memberlist does not change.

Expected: State of the Node in memberlist changes to a higher score, e.g.
```
Member: node8082 0
Member: node8081 1
```
...

PR #298, opened 2024-03-22
fix node state export

YuhangMa1117:
This commit fix the https://github.com/hashicorp/memberlist/issues/266 that get the confusion state of node when Members called, due to c192837f commit which involved two state varibale for one node.


dhilgarth:
Would this PR also fix this related issue? https://github.com/hashicorp/memberlist/issues/293

YuhangMa1117:
> Would this PR also fix this related issue? #293

I believe that #293 brings up a subject which is similar to mine, but the solutions he proposed may cause a considerable diff which is not so graceful. This PR will only fix the ```Members```  problem but cannot fix #293 because of the state is used in ```NotifyLeave```, and for this reason this PR is more clearly.

PR #301, opened 2024-06-14
simplify logic in queue's GetBroadcasts

mbver:
getting an item to broadcast based on priority of
 - transmits: the lower the better
 - msgLen: the longer the better
 - id: the higher the better
 starting from the front of queue check each item. if it is ok to be broadcasted. picked it and move on to the next. if it is not ok (length doesn't fit), still move on to the next. stop when there's no space for msgs or reach the end of queue.
for all items that are picked, first delete them from the queue. then increase its transmits. if the item's transmit is over transmitLitmit, finish it. Else insert it back to the queue for the next round of GetBroadcast.
The test is updated with variable message length and transmits.

### DIFF
--- a/queue.go
+++ b/queue.go
@@ -268,24 +268,6 @@ func (q *TransmitLimitedQueue) addItem(cur *limitedBroadcast) {
 	}
 }
 
-// getTransmitRange returns a pair of min/max values for transmit values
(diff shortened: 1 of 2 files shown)


mbver:
hi @rboyer , could you please give me some feedbacks when you have time? thank you.

PR #304, opened 2024-06-22
do not fire false ackMessage in timeout. check timeout in caller

mbver:
firing a false ackMessage at timeout will cause additional checks for v.Complete in the caller and further complicates handling by resending false ackMessage. instead, only check the timeout in the caller where it is needed. therefore, there's no need for the field Complete in ackMessage.


mbver:
hi @armon , could you please give a few comments on this when you have time?

PR #309, opened 2024-08-29
fix skip label check

reyreaud-l:
Memberlist allows having label and to check them on receiving a message to ensure rings don't collide unexpectedly. The option to skip that checks currently doesn't work and will fail the message unless the labels are exactly the same.

This PR change that behaviour to ensure that the option to skip the label check does that.

I tried adding test to showcase that the behaviour is now working as expected. Let me know if you any changes would be necessary. This is my first time contributing to that repository, let me know if I missed any necessary steps <3

### DIFF
--- a/net.go
+++ b/net.go
@@ -252,10 +252,6 @@ func (m *Memberlist) handleConn(conn net.Conn) {
 	}
 
 	if m.config.SkipInboundLabelCheck {
-		if streamLabel != "" {
-			m.logger.Printf("[ERR] memberlist: unexpected double stream label header: %s", LogConn(conn))
-			return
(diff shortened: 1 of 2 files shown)


rboyer:
This is working as intended.

`SkipInboundLabelCheck` is about loosening the requirement that inbound traffic be labeled at all, not that it be labeled correctly. This is expected to be set when the label has been received and removed by a proxy of sorts, before arriving at the memberlist where `SkipInboundLabelCheck` is set to `true`. Internally that latter memberlist is configured to know its label and must ensure that the protocol was not violated and accidentally two labels were sent.

PR #325, opened 2025-06-23
Upgradation of golangci-lint and linter fixes

KaushikiAnand:
Due to version upgradation of golangci-lint to version 2, linter was upgraded and few linter errors were fixed.


state.go
```diff
@@ -1231,7 +1228,7 @@ func (m *Memberlist) suspectNode(s *suspect) {
 
 		m.nodeLock.Lock()
 		state, ok := m.nodeMap[s.Node]
-		timeout := ok && state.State == StateSuspect && state.StateChange == changeTime
+		timeout := ok && state.State == StateSuspect && state.StateChange.Equal(changeTime)
```
abhijeetviswa, line 1231: Out of curiosity, what is the lint error here?

state_test.go
```diff
@@ -183,8 +207,13 @@ func TestMemberList_ProbeNode_Suspect_Dogpile(t *testing.T) {
...
 				})
-				defer peer.Shutdown()
+				defer func() {
+					if err := peer.Shutdown(); err != nil {
+						t.Fatal(err)
+					}
+				}()
 
+				//nolint:staticcheck // reason: peers used later indirectly
 				peers = append(peers, peer)
```
abhijeetviswa, line 217: `peers` is being appended to in this loop. But I don't see the array itself being used elsewhere. Where is it being used indirectly?

util.go
```diff
@@ -48,11 +51,9 @@ func decode(buf []byte, out interface{}) error {
...
 	buf := bytes.NewBuffer(nil)
 	buf.WriteByte(uint8(msgType))
-	hd := codec.MsgpackHandle{
-		BasicHandle: codec.BasicHandle{
-			TimeNotBuiltin: !msgpackUseNewTimeFormat,
-		},
-	}
+	hd := codec.MsgpackHandle{}
+	hd.TimeNotBuiltin = !msgpackUseNewTimeFormat
+
```
abhijeetviswa, line 56: Callout: No change in the wire format of hd, just removing a direct reference to `code.BasicHandle`. Similar to this [change](https://github.com/hashicorp/raft/pull/647/files/eec38fd04e67bba4484cbccdbdfd4cd4e5f595f7#r2153914642).

util.go
```diff
@@ -190,7 +191,9 @@ func makeCompoundMessage(msgs [][]byte) *bytes.Buffer {
 
 	// Add the message lengths
 	for _, m := range msgs {
-		binary.Write(buf, binary.BigEndian, uint16(len(m)))
+		if err := binary.Write(buf, binary.BigEndian, uint16(len(m))); err != nil {
+			log.Fatal(err)
+		}
```
abhijeetviswa, line 196: `log.Fatal` calls `os.Exit`, which is an extremely problematic side-effect.
Swallow the error instead.

util.go
```diff
@@ -33,7 +34,9 @@ const (
 )
 
 func init() {
-	seed.Init()
+	if _, err := seed.Init(); err != nil {
+		log.Fatal(err)
+	}
```
abhijeetviswa, line 39: log.Fatal calls os.Exit, which is an extremely problematic side-effect.
Swallow the error instead.

util.go
```diff
@@ -280,12 +283,16 @@ func decompressPayload(msg []byte) ([]byte, error) {
...
 	}
 
 	// Create a uncompressor
 	uncomp := lzw.NewReader(bytes.NewReader(c.Buf), lzw.LSB, lzwLitWidth)
-	defer uncomp.Close()
+	defer func() {
+		if err := uncomp.Close(); err != nil {
+			log.Fatal(err)
+		}
+	}()
```
abhijeetviswa, line 295: log.Fatal calls os.Exit, which is an extremely problematic side-effect.
Swallow the error instead.

abhijeetviswa (changes requested):

net_test.go
```diff
@@ -507,15 +555,13 @@ func TestTCPPushPull(t *testing.T) {
...
 	// Send our node state
 	header := pushPullHeader{Nodes: 3}
-	hd := codec.MsgpackHandle{
-		BasicHandle: codec.BasicHandle{
-			TimeNotBuiltin: !m.config.MsgpackUseNewTimeFormat,
-		},
-	}
+	hd := codec.MsgpackHandle{}
+	hd.TimeNotBuiltin = !m.config.MsgpackUseNewTimeFormat
+
```
abhijeetviswa, line 560: Callout: Similar to [here](https://github.com/hashicorp/memberlist/pull/325#discussion_r2161342333).

net.go
```diff
@@ -1359,6 +1369,7 @@ func (m *Memberlist) sendPingAndWaitForAck(a Address, ping ping, deadline time.T
 	}
 
 	if msgType != ackRespMsg {
+		//nolint: staticcheck
 		return false, fmt.Errorf("Unexpected msgType (%d) from ping %s", msgType, LogConn(conn))
```
abhijeetviswa, line 1373: What is the error/rule being ignored here? 
Also, let's remove the space after the colon (`:`).

memberlist.go
```diff
@@ -283,6 +283,7 @@ func (m *Memberlist) Join(existing []string) (int, error) {
 			hp := joinHostPort(addr.ip.String(), addr.port)
 			a := Address{Addr: hp, Name: addr.nodeName}
 			if err := m.pushPullNode(a, true); err != nil {
+				//nolint: staticcheck
 				err = fmt.Errorf("Failed to join %s: %v", a.Addr, err)
```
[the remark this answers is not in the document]
KaushikiAnand: memberlist.go:286:11: ST1005: error strings should not be capitalized (staticcheck)
                                err = fmt.Errorf("Failed to join %s: %v", a.Addr, err)
                                      ^

memberlist_test.go
```diff
@@ -788,7 +860,9 @@ func TestMemberlist_JoinDifferentNetworksMultiMasks(t *testing.T) {
 	c1.CIDRsAllowed, _ = ParseCIDRs([]string{"127.0.0.0/24", "127.0.1.0/24"})
 	m1, err := Create(c1)
 	require.NoError(t, err)
-	defer m1.Shutdown()
+	defer func() {
+		_ = m1.Shutdown()
+	}()
```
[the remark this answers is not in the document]
KaushikiAnand: Good catch. I missed it will correct it.

state.go
KaushikiAnand: state.go:1234:51: QF1009: probably want to use time.Time.Equal instead (staticcheck)
                timeout := ok && state.State == StateSuspect && state.StateChange == changeTime
                                                                ^

security.go
```diff
@@ -201,7 +198,7 @@ func decryptPayload(keys [][]byte, msg []byte, data []byte) ([]byte, error) {
 		}
 	}
 
-	//nolint: staticcheck
+	//nolint:staticcheck // ST1005: error strings should not be capitalized, reason: Test case failing
 	return nil, fmt.Errorf("No installed keys could decrypt the message")
```
abhijeetviswa, line 202: Let's fix the test case.

abhijeetviswa (changes requested):

security.go
KaushikiAnand: Resolved the comment in commit 5559c6e

memberlist.go
```diff
@@ -283,6 +283,7 @@ func (m *Memberlist) Join(existing []string) (int, error) {
 			hp := joinHostPort(addr.ip.String(), addr.port)
 			a := Address{Addr: hp, Name: addr.nodeName}
 			if err := m.pushPullNode(a, true); err != nil {
+				//nolint:staticcheck // ST1005: error strings should not be capitalized, reason: Test case failing
 				err = fmt.Errorf("Failed to join %s: %v", a.Addr, err)
```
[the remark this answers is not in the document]
KaushikiAnand: Resolved the comment in commit https://github.com/hashicorp/memberlist/commit/5559c6ef02bb8de5d3fdaf4459ee65b41dace9ba

PR #347, opened 2026-06-11
CI: fix broken YAML in config

tgross:
In https://github.com/hashicorp/memberlist/pull/346 I didn't notice that the tests weren't actually running, because the YAML of the CI config didn't parse.


gulducat (approved):
LGTM

PR #349, opened 2026-06-18
testing: reduce test flakiness

tgross:
Address various issues with test reliability.
* A number of the tests assumed that the cluster state diverged immediately, whereas we need to poll for it to settle.
* Ensure that port bindings for tests don't collide with ports in use on the host.
* Wait a little longer for push-pull metrics to be emitted.
* Fix some broken Make targets

Ref: https://hashicorp.atlassian.net/browse/NMD-1557

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,8 +18,8 @@ _cgo_gotypes.go
 _cgo_export.*
 
 _testmain.go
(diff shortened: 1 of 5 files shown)


gulducat (approved):
❄️

PR #350, opened 2026-06-19
fix probe selection for very small clusters

tgross:
When a node marks a peer as suspect, it measures its own health awareness by probing a subset of random nodes. The selection of k random nodes is via random offsets into the slice, but if n is very close to k and enough nodes are excluded by the filter function, it's possible to miss a node that we should be probing.

This scenario will be common in 3 node clusters as we commonly see with Nomad. If we're hitting this code path only 1 node will be non-excluded, so we expect to miss probing it ~2.6% of the time. This can result in false-negatives in health scoring, and also results in flaky tests of the lifeguard extension feature.

Add a new path for very small clusters that shuffles a copy of the nodes slice and iterates over it until we have k nodes or have exhausted the slice. Although this costs a bit extra for the shuffle, it also cuts the number of filter function calls from up to 3n to n, and eliminates the de-duplication check that iterates over the inprogress `kNodes` slice up to 3n times.

Also fixes a data race when setting up ack handlers, which was detected during testing. We shouldn't mutate the handler outside of the lock.

### How Has This Been Tested?

In addition to the test changes, I've run this with `go test -timeout=30m -count=100 -race . -run  TestMemberList_ProbeNode_Awareness` without failures.

### DIFF
--- a/state.go
+++ b/state.go
@@ -846,14 +846,8 @@ func (m *Memberlist) setProbeChannels(seqNo uint32, ackCh chan ackMessage, nackC
 		}
 	}
 
-	// Add the handlers
-	ah := &ackHandler{ackFn, nackFn, nil}
-	m.ackLock.Lock()
-	m.ackHandlers[seqNo] = ah
-	m.ackLock.Unlock()
-
-	// Setup a reaping routing
-	ah.timer = time.AfterFunc(timeout, func() {
+	// Add the handler, with a reaping routine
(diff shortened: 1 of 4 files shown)


jrasell (approved):
LGTM!

The repo has not been using its changelog; is now a good time to potentially start?

tgross:
> The repo has not been using its changelog; is now a good time to potentially start?

I was just going to remove it (I just did that on go-metrics too). These libraries change very slowly, so parsing thru a handful of commits in a diff seems like not a huge effort when you tag the library.

PR #357, opened 2026-07-07
limit remote state header values

tgross:
The header of a push-pull state message contains a count of nodes and user state length. As an optimization, both of these fields are used to pre-allocate memory to receive the remote state before copying it off the wire. But if the header mismatches the actual state, a very small pull-push state message (or at least one below the message size limit) can be used to consume excess memory on the receiver.

Limit the memory allocated such that the maximum amount allocated cannot exceed the maximum actual size of the push-pull state, either in number of nodes or size of the user state.

Ref: https://hashicorp.atlassian.net/browse/SECVULN-42161
Ref: https://hashicorp.atlassian.net/browse/NMD-1523

### DIFF
--- a/net.go
+++ b/net.go
@@ -86,7 +86,8 @@ const (
 	userMsgOverhead        = 1
 	blockingWarning        = 10 * time.Millisecond // Warn if a UDP packet takes this long to process
 	maxPushStateBytes      = 20 * 1024 * 1024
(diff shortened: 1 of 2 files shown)


jrasell (approved):

tgross:
Ref: https://discuss.hashicorp.com/t/hcsec-2026-18-memberlist-vulnerable-to-denial-of-service-via-gossip-message/77556/1

PR #359, opened 2026-07-07
reject implausibly high incarnation numbers

### LINKED ISSUE #358: refute() overflows to incarnation 0 when accusedInc is math.MaxUint32
refute()'s doc comment (state.go) says it "will make sure the incarnation number beats the given accusedInc value".

However in state.go:

```go
inc = m.skipIncarnation(accusedInc - inc + 1)   // atomic add on a uint32
```

If accusedInc is math.MaxUint32, this overflows and inc becomes 0.

Repro (add to the memberlist package, go test):

```go
  func TestRefuteOverflow(t *testing.T) {
        m := GetMemberlist(t, nil)
        defer m.Shutdown()

        m.aliveNode(&alive{Node: m.config.Name, Addr: []byte{127, 0, 0, 1}, Incarnation: 1, Vsn: m.config.BuildVsnArray()},
  true)
        m.suspectNode(&suspect{Node: m.config.Name, Incarnation: math.MaxUint32})

        if got := m.nodeMap[m.config.Name].Incarnation; got == 0 {
                t.Fatalf("refute wrapped our incarnation to 0")
        }
  }
```

Also in v0.6.0.

Not sure what the intended behavior is here.

Background: While experimenting with memberlist we found a single message can permanently evict a node.

Note: accusedInc comes straight off the wire, so any node that can inject a suspect message can trigger it.

Also at first glance I would say this is same class of issue as #357 fixed

Thanks.

Small update; have emailed security@hashicorp.com about this although its probably fine to have it here in public tracker, also made a PR #359 to address this.

joakimhellum:
## Description

A max incarnation in a suspect/dead/alive message overflows refute() and lets one message permanently evict a node. Ignore incarnations above a limit on ingestion, the same way #357 bounds the push-pull header.

## Related Issue

Fixes #358.

## How Has This Been Tested?

Full go test ./... passes

### DIFF
--- a/state.go
+++ b/state.go
@@ -951,6 +951,11 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 		return
 	}
 
+	// maxIncarnation is valid (the highest legit refutation); only above it is rejected.
+	if a.Incarnation > maxIncarnation {
+		return
+	}
+
 	if len(a.Vsn) >= 3 {
 		pMin := a.Vsn[0]
 		pMax := a.Vsn[1]
@@ -1089,6 +1094,11 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 
 	// If this is us we need to refute, otherwise re-broadcast
 	if !bootstrap && isLocalNode {
+		// A self-alive triggers refute(), so treat it as an accusation
+		if a.Incarnation >= maxIncarnation {
+			return
+		}
+
 		// Compute the version vector
 		versions := []uint8{
 			state.PMin, state.PMax, state.PCur,
@@ -1167,6 +1177,11 @@ func (m *Memberlist) suspectNode(s *suspect) {
 
 	// Ignore old incarnation numbers
 	if s.Incarnation < state.Incarnation {
+		return
+	}
+
+	// At/above the cap a refute would overshoot it, so reject (cf. > on alive).
+	if s.Incarnation >= maxIncarnation {
 		return
 	}
 
@@ -1260,6 +1275,11 @@ func (m *Memberlist) deadNode(d *dead) {
 
 	// Ignore old incarnation numbers
 	if d.Incarnation < state.Incarnation {
+		return
+	}
+
+	// At/above the cap a refute would overshoot it, so reject (cf. > on alive).
+	if d.Incarnation >= maxIncarnation {
 		return
 	}
 
--- a/state_test.go
+++ b/state_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"math"
 	"net"
 	"os"
 	"reflect"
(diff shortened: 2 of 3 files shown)


joakimhellum:
One note: a node at the max incarnation can't be failed. This is inherent to the counter, not something this PR fixes. Reaching it organically would take ~4 billion refutations, so a dead node would just stay in the member list instead of being removed.

state_test.go
```diff
@@ -2661,3 +2661,81 @@ func verifySampleExists(t *testing.T, name string, sink *metrics.InmemSink) {
...
+	m.aliveNode(&a, nil, false)
+
+	// An accusation at the ceiling is dropped, so the node can't be evicted: its
+	// refutation would leap to maxIncarnation+1 and be rejected as too high.
+	s := suspect{Node: "test", Incarnation: maxIncarnation}
+	m.suspectNode(&s)
+
+	if m.nodeMap["test"].State != StateAlive {
+		t.Fatalf("Bad state")
+	}
```
tgross: For all these tests, we can't read the `nodeMap` without taking a read lock, otherwise we can trigger data race detection. There's a `(*Memberlist).getNodeState` method you can use here.

Also, use [`testify`](https://pkg.go.dev/github.com/stretchr/testify) for assertions in this library.

state.go
```diff
@@ -1089,6 +1094,11 @@ func (m *Memberlist) aliveNode(a *alive, notify chan struct{}, bootstrap bool) {
 
 	// If this is us we need to refute, otherwise re-broadcast
 	if !bootstrap && isLocalNode {
+		// A self-alive triggers refute(), so treat it as an accusation
+		if a.Incarnation >= maxIncarnation {
+			return
+		}
```
tgross, line 1100: How would we hit this code path, given the change above?

tgross (changes requested):
The set of tests we've added only exercises that we're dropping the message, which is fairly trivial. The tests don't work out how the refuting node actually handles this state. And I think if you do that with `s := suspect{Node: "test", Incarnation: maxIncarnation - 1}` you'll find that we haven't actually fixed the bug, just set a cap on the incarnation so that it doesn't overflow anymore. We still end up with the victim node permanently out of the list, don't we? Do you have a test that exercises this?

Unfortunately we don't have a published security model for the library, but downstream applications like Nomad or Consul require using gossip encryption as part of their application-level securty model. Even then, a compromised member of the cluster could generate malicious messages: this is a byzantine consensus problem! It's not clear to me that memberlist is actually intended to be robust against byzantine members that aren't obeying the protocol.

We should fix the various DoS bugs you've pointed out in other issues/PRs from crashes, for sure, because those are simply bugs. But unless we can come up with an actual fix here I think this gets chalked up as a protocol gap?

joakimhellum:
Yeah, protocol gap was my initial thought, but please allow a few hours of research and testing here before I completely agree on it. Working on it now.

joakimhellum:
@tgross 

Updated the tests:
- Added one for the refuting node: a suspect at maxIncarnation-1 refutes to maxIncarnation and stays in the list.
- The suspect/dead tests use the local node, so the refute path runs, not just the drop.
- The aliveNode-at-maxIncarnation test hits the state.go path you asked.

Also this is my initial opinion:

I think capping incarnation on ingest might actually be the right step, because its true to SWIM, and other systems do similar. For example:

https://github.com/apple/swift-cluster-membership/blob/main/Sources/SWIM/SWIMInstance.swift#L1568-L1574

https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/gms/Gossiper.java#L1494-L1498

Both treat "a remote supplied counter that is too far ahead" as invalid input to ignore. Cassandra don't directly translate, though. SWIM's incarnation/refute exists so a healthy node can always overrule a false suspicion, so the design leans toward keeping healthy nodes in the cluster, which is what the overflow breaks here.

So to me it's a trade off between two failure modes; wrongly evicting healthy node vs. failing to remove a specific node 

Although the implementation here might not be right.

I do have some suggestions, but I need some more time to research and test, or else you get suggestions that far exceeds what you would be comfy with in this PR.

Update 1: This PR suggests absolute cap, what I'm currently testing is cap + bound the leap (which probably far exceeds what this PR should do already). Will report when tested.

Update 2: Seems like cap + relative Δ preserves restarts and closes the weird "1 packet pinning" here. But not sure I am comfy with it. As you said, Nomad and Consul use this. While the full test suite passes on it, it's just me guessing. Also it is mitigation, it won't completely close the issue.

Update 3: Thinking about it more, I would split these. The overflow on its own is just a bug: refute wraps to the lowest incarnation instead of the highest, and the cap fixes it. Then relative Δ only softens it but comes with some sizing questions like we already have.

Update 4: To be more correct here; the cap converts "attacker evicts a healthy node" into "attacker pins a node unremovable + meta-frozen". Which to me is milder and bounded failure mode. And cap + Δ might soften that in a future PR.

I am looking for some guidance on this honestly.

For context: we experiment with memberlist under signed node identity, so ingress hardening like this matters to us (which is why we went down this route in the first place).

joakimhellum:
@tgross

I have gone back and forth here many times now, and touching failure semantics of Consul and Nomad makes me worried. But the alternative to this fix is leaving the overflow in, and after thinking about this over and over, I think the overflow is worse for Consul/Nomad than the pinning. 

Both the overflow and the pinning need an "attacker" and the difference is what is left. The overflow wraps a healthy node to ~0 and neither refute nor push/pull can recover it, while the cap just pins the node, so it stays a working member unless it later actually dies. So leaving the overflow in is the worse option for Consul/Nomad, so I still think fixing this and treating softening the pin (cap+Δ) as a follow-up might be the right approach.

However all this thinking is probably a sign that many eyes should be on it before any decision is made.

Thanks for reading through all of this.

Update: every guard should be tested now. In addition it also tests:
- A suspect at maxIncarnation-1 makes the node refute to maxIncarnation and stay alive (TestMemberList_SuspectNode_NearMaxIncarnation)
- A node parked at the ceiling can't be failed later (TestMemberList_PinnedNodeIsUnfailable)

A note about refute(): its job is to push incarnation above the accusation so a healthy node overrules a false suspicion, but near MaxUint32 the +1 wraps, so it lands below the accusation instead and the node loses its own refute. So the function quietly breaks its own postcondition on out of range incarnation. The existing refute mechanism just stops working at the top of the range. So rejecting out of range numeric field on ingest instead of letting it drive the counter does sound right in my opinion, because it's same class as #357 in this context.

joakimhellum:
Added a new test for another residual (TestMemberList_CounterDriftsPastCap): the cap is enforced on accepted messages, not on the counter. When a node at the ceiling refreshes, its incarnation climbs past maxIncarnation and the message gets dropped. Clamping the counter would probably fix it, but not sure it's worth the effort here. Should be clear by now this is quite complicated stuff. I'm happy to continue here if @tgross sees any value in it.

Example clamp in `skipIncarnation`:

```go
// nextIncarnation returns the next incarnation number in a thread safe way
func (m *Memberlist) nextIncarnation() uint32 {
	return m.skipIncarnation(1)
}

// skipIncarnation adds the positive offset to the incarnation number, saturating
// at maxIncarnation so the counter never runs past the cap enforced on the wire.
func (m *Memberlist) skipIncarnation(offset uint32) uint32 {
	for {
		cur := m.incarnation.Load()
		next := cur + offset
		if next < cur || next > maxIncarnation { // wrapped, or past the cap
			next = maxIncarnation
		}
		if m.incarnation.CompareAndSwap(cur, next) {
			return next
		}
	}
}
```

My final opinion: This deserves discussion but probably shouldn't merge as-is. Normally trying to understand a bug makes the fix clearer, but in this case its the opposite, so taking it as a sign.

tgross:
> My final opinion: This deserves discussion but probably shouldn't merge as-is. Normally trying to understand a bug makes the fix clearer, but in this case its the opposite, so taking it as a sign.

Yeah, I'm going to poke some folks here internally as well to see if they have thoughts.

joakimhellum:
@tgross Since a [security model](https://github.com/hashicorp/memberlist/blob/master/SECURITY.md#memberlist-security-model) has been published, I am planning to close this issue and PR. There is a list of stuff reported through the mailing list, and some of these are simple bugs, which I am happy to make PRs for. But others (like this issue) turned out to be very complicated and have to be considered "protocol gaps" which is now covered by the new [security model](https://github.com/hashicorp/memberlist/blob/master/SECURITY.md#memberlist-security-model) docs.

joakimhellum:
For anyone reading this for any reason, please refer to the [security model](https://github.com/hashicorp/memberlist/blob/master/SECURITY.md#memberlist-security-model) and specifically the [trust model](https://github.com/hashicorp/memberlist/blob/master/SECURITY.md#trust-model). 

This PR is very complicated and would require at least 2 additional follow up PRs to close out the residuals found **so far**, which is reason why I am not comfy leaving it here open at all. And it feels pointless with so much pain since this is now covered by the [security model](https://github.com/hashicorp/memberlist/blob/master/SECURITY.md#memberlist-security-model).

tgross:
I meant to come back and address this once the security model PR got merged but got side-tracked. Thanks for the effort here @joakimhellum !

PR #361, opened 2026-07-08
limit user message length

### LINKED ISSUE #360: readUserMsg: unbounded make([]byte, header.UserMsgLen), same class as #357
`readUserMsg` (net.go) allocates `make([]byte, header.UserMsgLen)` straight from the wire header, with no bound. A small `userMsg` frame claiming a huge length forces the allocation up front. #357 added this bound to `readRemoteState` (net.go); `readUserMsg` might have been missed.

```go
  func TestUserMsgUnboundedAlloc(t *testing.T) {
        const claimed = 512 << 20 // header claims 512 MiB; no body is sent

        hd := codec.MsgpackHandle{}
        var hdr bytes.Buffer
        if err := codec.NewEncoder(&hdr, &hd).Encode(&userMsgHeader{UserMsgLen: claimed}); err != nil {
                t.Fatal(err)
        }
        m := &Memberlist{config: DefaultLANConfig()}
        stream := bytes.NewReader(hdr.Bytes())
        dec := codec.NewDecoder(stream, &hd)

        var before, after runtime.MemStats
        runtime.GC()
        runtime.ReadMemStats(&before)
        err := m.readUserMsg(stream, dec) // fails on the absent body...
        runtime.ReadMemStats(&after)

        // ...only after allocating the full claimed size from the header alone.
        t.Logf("%d-byte header, UserMsgLen=%d -> allocated %d MiB (err=%v)",
                hdr.Len(), claimed, (after.TotalAlloc-before.TotalAlloc)>>20, err)
        if after.TotalAlloc-before.TotalAlloc < claimed/2 {
                t.Fatal("allocation not bounded")
        }
  }
```

`17-byte header, UserMsgLen=536870912 -> allocated 512 MiB (err=EOF)`

joakimhellum:
## Description

readUserMsg sizes make([]byte, header.UserMsgLen) from the wire header with no bound, so a small userMsg frame declaring a large length forces a large allocation on the receiver. This is the userMsg path of the same issue #357 fixed for push-pull state.

## Related Issue

Fixes #360 

## How Has This Been Tested?

Full go test ./... passes

### DIFF
--- a/net.go
+++ b/net.go
@@ -86,8 +86,9 @@ const (
 	userMsgOverhead        = 1
 	blockingWarning        = 10 * time.Millisecond // Warn if a UDP packet takes this long to process
 	maxPushStateBytes      = 20 * 1024 * 1024
-	maxPushStateNodes      = 1024 * 1024 // Each requires conservatively  ~20 bytes when encoded
-	maxPushPullRequests    = 128         // Maximum number of concurrent push/pull requests
+	maxPushStateNodes      = 1024 * 1024      // Each requires conservatively  ~20 bytes when encoded
+	maxUserMsgBytes        = 20 * 1024 * 1024 // Largest user message we will buffer off the wire
+	maxPushPullRequests    = 128              // Maximum number of concurrent push/pull requests
 )
 
 // ping request sent directly to node
@@ -1323,6 +1324,10 @@ func (m *Memberlist) readUserMsg(bufConn io.Reader, dec *codec.Decoder) error {
 	var header userMsgHeader
 	if err := dec.Decode(&header); err != nil {
 		return err
+	}
+
+	if header.UserMsgLen < 0 || header.UserMsgLen > maxUserMsgBytes {
+		return fmt.Errorf("user message length (%d) exceeds limit", header.UserMsgLen)
 	}
 
 	// Read the user message into a buffer
--- a/net_test.go
+++ b/net_test.go
@@ -1074,6 +1074,42 @@ func TestReadRemoteState_Limits(t *testing.T) {
 	})
 }
 
+func TestReadUserMsg_Limit(t *testing.T) {
+
+	mockNet := &MockNetwork{}
+	tr := mockNet.NewTransport("node")
+	logs := &bytes.Buffer{}
+	logger := log.New(logs, "", 0)
+
+	m := GetMemberlist(t, func(c *Config) {
+		c.EnableCompression = false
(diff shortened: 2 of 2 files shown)


joakimhellum:
To be clear: a single userMsg header with an oversized length crashes the receiver (unrecovered panic / fatal OOM), so it is not just memory pressure.

net.go
```diff
@@ -86,8 +86,9 @@ const (
 	userMsgOverhead        = 1
 	blockingWarning        = 10 * time.Millisecond // Warn if a UDP packet takes this long to process
 	maxPushStateBytes      = 20 * 1024 * 1024
-	maxPushStateNodes      = 1024 * 1024 // Each requires conservatively  ~20 bytes when encoded
-	maxPushPullRequests    = 128         // Maximum number of concurrent push/pull requests
+	maxUserMsgBytes        = 20 * 1024 * 1024 // Largest user message we will buffer off the wire
+	maxPushStateNodes      = 1024 * 1024      // Each requires conservatively  ~20 bytes when encoded
+	maxPushPullRequests    = 128              // Maximum number of concurrent push/pull requests
```
tgross: Super nitpick: can we group the push state and the user msg constants together?

net_test.go
```diff
@@ -1074,6 +1074,42 @@ func TestReadRemoteState_Limits(t *testing.T) {
...
+	conn, err := tr.DialTimeout(addr, time.Millisecond*100)
+	require.NoError(t, err)
+
+	err = m.rawSendMsgStream(conn, buf.Bytes(), "")
+	require.NoError(t, err)
+
+	// conn closed: get nothing back
+	var out []byte
+	_, err = conn.Read(out)
+	require.Error(t, err, "EOF")
```
tgross: I think we mean:

```suggestion
	require.ErrorContains(t, err, "EOF")
```

`require.Error` is just that we get an error, and `"EOF"` is the message returned to the developer if we don't get one. Bit of a footgun in testify's API (and one of the many things fixed in [`shoenig/test`](https://pkg.go.dev/github.com/shoenig/test), which we use on Nomad but not this library).

tgross (changes requested):
I've left a couple minor suggestions but otherwise this LGTM

joakimhellum:
Thanks. Constants grouped and require.ErrorContains fixed

tgross (approved):
LGTM. Thanks @joakimhellum!

PR #363, opened 2026-07-08
limit decompressed message size

### LINKED ISSUE #362: decompressBuffer expands an LZW compressMsg with no bound
`decompressBuffer` (util.go) decompresses a `compressMsg` off the wire into a buffer with `io.Copy`, no bound. A small compressMsg expands to many times its size. The remote-state copy is bounded with `io.CopyN` (net.go); the decompress copy isn't.

```go
  func TestDecompressUnbounded(t *testing.T) {
        // A small compressMsg carrying 256 MiB of zeros, LZW-compressed.
        var cbuf bytes.Buffer
        w := lzw.NewWriter(&cbuf, lzw.LSB, lzwLitWidth)
        if _, err := w.Write(make([]byte, 256<<20)); err != nil {
                t.Fatal(err)
        }
        if err := w.Close(); err != nil {
                t.Fatal(err)
        }

        out, err := decompressBuffer(&compress{Algo: lzwAlgo, Buf: cbuf.Bytes()})
        if err != nil {
                t.Fatal(err)
        }
        t.Logf("%d-byte compressMsg -> %d MiB decompressed (%dx), no cap",
                cbuf.Len(), len(out)>>20, len(out)/cbuf.Len())
        if len(out) < 100<<20 {
                t.Fatalf("expected a large decompressed buffer, got %d bytes", len(out))
        }
  }
```
 
`198106-byte compressMsg -> 256 MiB decompressed (1355x), no cap`

 Impact: a single compressMsg can be inflated until the receiver runs out of memory.

I reported two other issues on back of #357 , which I still don't have any feedback on, so I will hold any PR for this until I have feedback, to not pollute your issue tracker and PRs.

Small update; have made a PR for this because I'm going on vacation soon.

joakimhellum:
## Description

decompressBuffer drains the LZW reader with io.Copy into an unbounded buffer, so a small compressMsg expands without limit and can exhaust memory. Cap the output at maxDecompressedBytes with io.CopyN and error past it.

## Related Issue

Fixes #362 

## How Has This Been Tested?

Full go test ./... passes

### DIFF
--- a/net.go
+++ b/net.go
@@ -89,6 +89,8 @@ const (
 	maxPushStateNodes      = 1024 * 1024      // Each requires conservatively  ~20 bytes when encoded
 	maxUserMsgBytes        = 20 * 1024 * 1024 // Largest user message we will buffer off the wire
 	maxPushPullRequests    = 128              // Maximum number of concurrent push/pull requests
+
+	maxDecompressedBytes = 2 * maxPushStateBytes // Largest push/pull we will decompress: user state plus an equal node budget
 )
 
 // ping request sent directly to node
--- a/util_test.go
+++ b/util_test.go
@@ -399,3 +399,19 @@ func TestCompressDecompressPayload(t *testing.T) {
 		t.Fatalf("bad payload: %v", decomp)
 	}
 }
+
+func TestDecompressPayload_Limit(t *testing.T) {
+	// Over the limit: rejected.
+	buf, err := compressPayload(make([]byte, maxDecompressedBytes+1), false)
+	require.NoError(t, err)
+
+	_, err = decompressPayload(buf.Bytes()[1:])
+	require.ErrorContains(t, err, "larger than limit")
+
+	// Between the wire limit and the decompressed limit: accepted.
+	buf, err = compressPayload(make([]byte, maxPushStateBytes+1), false)
+	require.NoError(t, err)
+
+	_, err = decompressPayload(buf.Bytes()[1:])
+	require.NoError(t, err)
+}
--- a/util.go
+++ b/util.go
@@ -302,11 +302,14 @@ func decompressBuffer(c *compress) ([]byte, error) {
 		_ = uncomp.Close()
 	}()
 
-	// Read all the data
+	// Ensure a compressMsg can't decompress to too much data.
 	var b bytes.Buffer
-	_, err := io.Copy(&b, uncomp)
-	if err != nil {
+	n, err := io.CopyN(&b, uncomp, maxDecompressedBytes+1)
+	if err != nil && err != io.EOF {
 		return nil, err
+	}
+	if n > maxDecompressedBytes {
+		return nil, fmt.Errorf("decompressed message is larger than limit (%d)", maxDecompressedBytes)
 	}
 
 	// Return the uncompressed bytes


util_test.go
```diff
@@ -399,3 +399,20 @@ func TestCompressDecompressPayload(t *testing.T) {
...
 }
+
+func TestDecompressPayload_Limit(t *testing.T) {
+	// Over the limit: rejected.
+	buf, err := compressPayload(make([]byte, maxDecompressedBytes+1), false)
+	require.NoError(t, err)
+
+	_, err = decompressPayload(buf.Bytes()[1:])
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "larger than limit")
```
tgross:
```suggestion
	require.ErrorContains(t, err, "larger than limit")
```

net.go
```diff
@@ -86,8 +86,9 @@ const (
 	userMsgOverhead        = 1
 	blockingWarning        = 10 * time.Millisecond // Warn if a UDP packet takes this long to process
 	maxPushStateBytes      = 20 * 1024 * 1024
-	maxPushStateNodes      = 1024 * 1024 // Each requires conservatively  ~20 bytes when encoded
-	maxPushPullRequests    = 128         // Maximum number of concurrent push/pull requests
+	maxDecompressedBytes   = 2 * maxPushStateBytes // push-pull: user state plus an equal node budget
```
tgross: Can you provide a little more motivation on this specific size? We compress first then encrypt, so we'd be decrypting first and then decompressing. So the max decompressed size needs to cover the size of the maximum _compound_ message size, doesn't it?

(Also, super nitpick: let's order these to group the push state constants separately from the new constants we're adding)

tgross (commented):
Thanks for the PR @joakimhellum!

The overall approach here looks good, but I'd like to see if we can dial-in the specific constant we've chosen here as the maximum size.

tgross (changes requested):

joakimhellum:
> Thanks for the PR @joakimhellum!
> 
> The overall approach here looks good, but I'd like to see if we can dial-in the specific constant we've chosen here as the maximum size.

Yeah, this needs a bit of research and testing/measuring. Working on it.

joakimhellum:
@tgross Compound messages are ~1.2KB normally (bounded by UDPBufferSize=1400) and ~65KB at most. The big thing that gets decompressed is push/pull state, which is what the 40MB is for; the 20MB user-state cap plus room for the node list on top. But I'm not really sure about the max cluster size. maxPushStateNodes says 1M, Consul says ~5-8K (?), and Nomad don't recommend too many per region. In practice it covers around ~50K with full Meta. 40MB is really only meant to cover "most realistic scenarios with some room to spare", while keeping the decompression bounded. Supporting 1M here would defeat the point of doing it in first place. So looking for a bit of guidance.

Clarification 1: push/pull goes through sendLocalState -> rawSendMsgStream -> decompressBuffer, which is the big thing;  the 20MB UserStateLen cap from #357 plus ~20MB for the node list

Clarification 2: I didn't find a way to derive a max size here. Name has no length limit and MetaMaxSize is only on send, so merge/receive has no check. So 40MB is more of a policy choice.

For self-reference: decompressed payload is header + node list + user state. a pushPullHeader (Nodes, UserStateLen) is a few bytes. header.Nodes x pushNodeState is the node list. the whole thing is what rawSendMsgStream compresses and decompressBuffer inflates.  2 * maxPushStateBytes breakdown is 20M user state cap (which is hard cap), and then 20MB node list budget, which is maxPushStateNodes (1M) × ~20 bytes/node (fetched from #357).  packet path is rawSendMsgPacket compresses compound message (UDP) -> handleCompressed -> decompressBuffer. stream path is sendLocalState -> rawSendMsgStream compress push/pull state (TCP) -> readSteam -> decompressBuffer, which is up to 40MB.

tgross:
Ok, the reasoning there seems sound: the maximum sizes we've set on the user msg / node list are more than what we'd expect from a "reasonable" size, and you've derived the max decompressed value from there, which seems reasonable. It occurs to me that a better design would have been to stream the compression into the decoder and then we could apply the limit once where it's being consumed. But that's a big refactor at this point.

Looks like there's some merge conflicts from my merging-in of #361. Once those are resolved we can merge this in.

joakimhellum:
Rebase done. Tests still green.

tgross (approved):
LGTM! Thanks @joakimhellum

PR #366, opened 2026-07-21
docs: clarify SendBestEffort does not enforce UDPBufferSize

### LINKED ISSUE #348: UDPBufferSize is not enforced in SendBestEffort despite its docs
The docs of `SendBestEffort` state that "The maximum size of the message depends on the configured UDPBufferSize for this memberlist instance":
https://github.com/hashicorp/memberlist/blob/29437a621ed5ed7975fd156c5053076828246e44/memberlist.go#L582-L595

I understood this meant I could not send a message larger than `UDPBufferSize` bytes using `SendBestEffort`, but this is not the case. Nowhere in the `SendBestEffort` call chain (`SendBestEffort` -> `rawSendMsgPacket` -> `NodeAwareTransport.WriteToAddress`) is `UDPBufferSize` checked nor mentioned.

Where the library uses it is in `sendMsg` and `gossip`: both take it into account to "fill a UDP packet with piggybacked data" via `getBroadcast`.

Should the comment in `SendBestEffort`  be removed, since it does not use it?

Solaris-star:
## Description

`SendBestEffort` docs said the max message size depends on `Config.UDPBufferSize`. That is not enforced on this path.

Call chain: `SendBestEffort` → `rawSendMsgPacket` → transport `WriteToAddress` — no `UDPBufferSize` check.

`UDPBufferSize` is used when packing piggybacked gossip (`sendMsg` / `gossip` via `getBroadcast`).

## Fix

Document actual behavior:

- no hard max from `UDPBufferSize` on `SendBestEffort`
- oversized messages may fail at transport/OS UDP layer
- use `SendReliable` for large messages
- `UDPBufferSize` still matters for gossip packing

## Tests

Docs-only. `go test -c` compiles.

## Linked Issue

Closes #348

### DIFF
--- a/memberlist.go
+++ b/memberlist.go
@@ -581,8 +581,12 @@ func (m *Memberlist) SendToTCP(to *Node, msg []byte) error {
 
 // SendBestEffort uses the unreliable packet-oriented interface of the transport
 // to target a user message at the given node (this does not use the gossip
-// mechanism). The maximum size of the message depends on the configured
-// UDPBufferSize for this memberlist instance.
+// mechanism).
+//
+// Unlike gossip/sendMsg filling, SendBestEffort does not enforce Config.UDPBufferSize
+// as a maximum payload size; oversized messages may still be written and can fail
+// at the transport or OS UDP layer. Prefer SendReliable for large messages.
+// UDPBufferSize is used when packing piggybacked gossip into UDP packets.
 func (m *Memberlist) SendBestEffort(to *Node, msg []byte) error {
 	// Encode as a user message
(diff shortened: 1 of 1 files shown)


tgross:
@Solaris-star slow down. We're simply don't have the bandwidth for you to be spraying our repositories with LLM-generated low-effort PRs, and aren't interested in reviewing a giant pile of them so you can goose your GitHub stats.

*Do not* open another PR until we've reviewed the ones you've made.

tgross:
This is the wrong fix. Consider that you probably shouldn't open PRs on issues that haven't actually been validated by maintainers.

PR #368, opened 2026-07-23
fix panic in Vsn deference

tgross:
The protocol version array is not length-checked everywhere, which can result in a panic if we deference a Vsn off the wire that's short.

Ref: https://hashicorp.atlassian.net/browse/PSA-2897
Ref: https://hashicorp.atlassian.net/browse/NMD-1624

### DIFF
--- a/net.go
+++ b/net.go
@@ -1291,6 +1291,10 @@ func (m *Memberlist) mergeRemoteState(join bool, remoteNodes []pushNodeState, us
 	if join && m.config.Merge != nil {
 		nodes := make([]*Node, len(remoteNodes))
 		for idx, n := range remoteNodes {
+			// Skip nodes without proper version information
+			if len(n.Vsn) < 6 {
+				continue
+			}
 			nodes[idx] = &Node{
 				Name:  n.Name,
 				Addr:  n.Addr,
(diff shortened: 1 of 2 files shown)


net.go
```diff
@@ -1291,6 +1291,10 @@ func (m *Memberlist) mergeRemoteState(join bool, remoteNodes []pushNodeState, us
 	if join && m.config.Merge != nil {
 		nodes := make([]*Node, len(remoteNodes))
 		for idx, n := range remoteNodes {
+			// Skip nodes without proper version information
+			if len(n.Vsn) < 6 {
+				continue
+			}
```
chrisroberts: Would it makes sense to "stub" the node in here instead of skipping entirely, just not filling in the `Vsn` related data? Since this `nodes` value gets passed on to `NotifyMerge`, if the implementation is iterating and assuming the slice contains only valid nodes, it may end up hitting a `nil` unexpectedly.

chrisroberts (commented):
Looks good, but have a suggestion/question as I start to dig into this library more.

net.go
tgross: You're right. This is only for the delegate notification and the original `pushNodeState` object ends up getting sent to all the downstream code anyways, whether or not `Vsn` is the correct length. So I've changed this to create the `*Node` unconditionally and only add the `Vsn` info if it's the correct length.

chrisroberts (approved):
LGTM

PR #369, opened 2026-07-23
check length of decompressed buffer before indexing

tgross:
If a compressed message is empty after decompression, the `readStream` method will attempt to index into it and panic. Do a length check here and treat such malformed messages as an error.

Ref: https://hashicorp.atlassian.net/browse/PSA-2897
Ref: https://hashicorp.atlassian.net/browse/NMD-1624

### DIFF
--- a/net.go
+++ b/net.go
@@ -7,6 +7,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"hash/crc32"
 	"io"
@@ -1216,7 +1217,7 @@ func (m *Memberlist) readStream(conn net.Conn, streamLabel string) (messageType,
 			return 0, nil, nil, err
 		}
 		if len(decomp) == 0 {
-			return 0, nil, nil, fmt.Errorf("decompressed message is empty")
+			return 0, nil, nil, errors.New("decompressed message is empty")
 		}
 
 		// Reset the message type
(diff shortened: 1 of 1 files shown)


chrisroberts (dismissed):

net.go
```diff
@@ -1215,6 +1215,9 @@ func (m *Memberlist) readStream(conn net.Conn, streamLabel string) (messageType,
 		if err != nil {
 			return 0, nil, nil, err
 		}
+		if len(decomp) == 0 {
+			return 0, nil, nil, fmt.Errorf("decompressed message is empty")
```
jrasell: nit and totally optional: we are not formatting the error, so could use a non-formatting func. The file does not import `errors`, so the suggestion alone doesn't work.
```suggestion
			return 0, nil, nil, errors.New("decompressed message is empty")
```

jrasell (dismissed):
LGTM; tiny inline suggestion but totally ignorable.

jrasell (approved):

net.go
tgross: Fixed. Note that for applications built with 1.26 (and for this library once Go 1.25 is out of support), these are equivalent because it's been `fmt.Errorf` has been optimized for this case so that we don't have to switch between the two during editing anymore :grinning: The CL has pretty good details on this https://go-review.googlesource.com/c/go/+/708836

PR #376, opened 2026-09-05
Return node copies from Members to avoid data races

### LINKED ISSUE #306: Data race in state.go:1135
Hi. I've detected the following data race in memberlist 0.5.0:

```
==================
WARNING: DATA RACE
Write at 0x00c000693728 by goroutine 53:
  github.com/hashicorp/memberlist.(*Memberlist).aliveNode()
      .../vendor/github.com/hashicorp/memberlist/state.go:1135 +0x2e3b
  github.com/hashicorp/memberlist.(*Memberlist).handleAlive()
      .../vendor/github.com/hashicorp/memberlist/net.go:738 +0x69a
  github.com/hashicorp/memberlist.(*Memberlist).packetHandler()
      .../vendor/github.com/hashicorp/memberlist/net.go:513 +0x1b9
  github.com/hashicorp/memberlist.newMemberlist.gowrap3()
      .../vendor/github.com/hashicorp/memberlist/memberlist.go:235 +0x33

Previous read at 0x00c000693728 by goroutine 146:
  github.com/hashicorp/memberlist.(*Node).Address()
      .../vendor/github.com/hashicorp/memberlist/state.go:58 +0x6b
...

Goroutine 53 (running) created at:
  github.com/hashicorp/memberlist.newMemberlist()
      .../vendor/github.com/hashicorp/memberlist/memberlist.go:235 +0x1aea
  github.com/hashicorp/memberlist.Create()
      .../vendor/github.com/hashicorp/memberlist/memberlist.go:246 +0x2e
...

Goroutine 146 (running) created at:
...
==================
```

The idea is that memberlist `Members()` returns a list of pointers to `Memberlist::nodes` array items. Then the external user's code is allowed to read it without any locking. But at the same time, it may get modified internally in Memberlist, resulting in a data race.

Not sure if it's a critical issue and how hard it would be to stumble on it in production. But it's still a race...

BetterAndBetterII:
## Summary

`Members()` returned pointers into the internal node list. Callers are expected to read those nodes without holding `nodeLock`, while `aliveNode` can update `Addr`/`Port`/`Meta` concurrently. That is the data race reported in #306 (`aliveNode` write vs `Node.Address()` read).

This change returns a copy of each alive `Node` so those reads no longer share storage with the live membership table.

## Test plan

- [x] `go test -count=1 -race -run 'TestMemberList_Members|TestMembers_ReturnsIndependentCopy|TestMembers_NoRaceWithAliveNode' .`
- [x] `go test -count=1 .`

Fixes #306


tgross:
@BetterAndBetterII you have made _far_ too many PRs across multiple library repos in a short span. Wait until you've received reviews on all the ones you've made so far before making another to avoid being classified as spam.

We'll review these as we have time.

